PR #13, opened 2024-07-01
Adding Rust implementation of Alba

rrtoledo:
This PR aims at adding the Rust implementation of Alba. This includes:
- Prehashed version & Bounded version: some TODO left to fully implement the recursive hash optimization and look at hash personalization, etc.
- (limited) tests:  positive and negative tests. Need to add randomized tests as well as property testing.
- (limited) doc: Succinct comments on code that can be compiled with cargo doc
- (limited) bench:  bench in tests (add assert!(false) to display) with low parameters, need to separate these in distinct file.


abailly-iohk:
I think we should add Rust CI along with this PR, at a minimum.

curiecrypt:
> I think we should add Rust CI along with this PR, at a minimum.

@abailly-iohk I was thinking the same thing. I am going to work on that 👍

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    /// Oracle producing a uniformly random value in [1, n_p] used for prehashing S_p
+    // TODO: We also return hash(data) to follow the optimization presented in Section 3.3
+    fn h0(setup: &Setup, v: usize, s: [u8; 32]) -> usize {
+        let mut data = Vec::new();
+        data.push(v.to_ne_bytes().to_vec());
+        data.push(s.to_vec());
+        let digest = utils::combine_hashes(data);
+        // return (digest, utils::oracle(&digest, setup.n_p));
```
abailly-iohk, line 189: Remove commented code

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// and returns first round candidate Round{t, x_1, ..., x_u)} such that:
+    /// - for all i ∈ [0, u-1], H0(x_i+1) ∈ bins[H1(t, x_1, ..., x_i)]
+    /// - H2(t, x_0, ..., x_u) = true
+    fn dfs(
+        setup: &Setup,
+        bins: &Vec<Vec<[u8; 32]>>,
+        round: &Round,
+        limit: &mut usize,
+    ) -> Option<Proof> {
+        // If the round candidate is at the last iteration (len(round) == u)
```
abailly-iohk, line 215: Comments seem redundant here, as they are merely paraphrasing the code just below

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+        if round.s_list.len() == setup.u {
+            // if H2(t, x_0, ..., x_u) = true, return candidate as proof
```
abailly-iohk, line 217: ditto

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+            if Proof::h2(setup, round) {
+                let r = round.v;
+                let d = round.t;
+                let items = round.s_list.clone();
+                return Some(Proof { r, d, items });
+            } else {
+                return None;
+            }
+        }
+        // Otherwise, update round candidate with all s in bins[H1(t, x_1, ..., x_i)] and continue
```
abailly-iohk, line 227: I would remove the comment here too. `Otherwise` seems to imply we do this _only_ if we don't do the other thing, which is not th case.

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Returns an empty proof if no suitable candidate is found.
+    fn prove_index(setup: &Setup, set: &Vec<[u8; 32]>, v: usize) -> (usize, Option<Proof>) {
+        // Initialising our n_p bins with all s ∈ Sp
+        let mut bins: Vec<Vec<[u8; 32]>> = Vec::new();
+        for _ in 1..(setup.n_p + 1) {
+            bins.push(Vec::new());
+        }
+        for &s in set.iter() {
+            // TODO: add H1(s),
+            // let (hs, index_s) = Proof::h0(setup, s);
```
abailly-iohk, line 248: Remove commented code (plus the `TODO` is misleading as it mentions `H1` and we use `h0`

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+            }
+            limit = limit - 1;
+            let round = Round::new(v, t, setup.n_p);
+            let res = Proof::dfs(setup, &bins, &round, &mut limit);
+            if res.is_some() {
+                return (limit, res);
+            }
+        }
+        // If no proof found, return None
+        return (limit, None);
```
abailly-iohk, line 266: The `limit` name is a bit counterintuitive as we _decrement_ it on each turn. Also, its not documented in the comment for the function.

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    /// Alba's proving algorithm, based on a depth-first search algorithm.
```
abailly-iohk, line 269: This comment is the same as the previous function, which might be confusing for the prospective user approaching the library for the first time

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Returns an empty proof if no suitable candidate is found.
+    pub fn prove(setup: &Setup, set: &Vec<[u8; 32]>) -> Self {
+        // Initialising our n_p bins with all s ∈ Sp
```
abailly-iohk, line 272: I would remove the comment here

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+        for v in 0..setup.r {
+            if let (_, Some(proof)) = Proof::prove_index(setup, set, v) {
+                return proof;
+            }
+        }
+        // If no proof found, return empty proof
```
abailly-iohk, line 278: ditto

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+        return Proof::empty();
+    }
+
+    /// Alba's proving algorithm, based on a depth-first search algorithm.
```
abailly-iohk, line 282: Another instance of the same comment, sound like a copy/pasting gone rogue :-D

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Returns an empty proof if no suitable candidate is found.
+    pub fn bench(setup: &Setup, set: &Vec<[u8; 32]>) -> (usize, Self) {
+        // Initialising our n_p bins with all s ∈ Sp
```
abailly-iohk, line 285: Remove comment

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+        let mut nb_calls = 0;
+        for v in 0..setup.r {
+            let (steps, opt) = Proof::prove_index(setup, set, v);
+            // Adding nb of steps done in prove_index
```
abailly-iohk, line 289: ditto

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+            nb_calls += setup.b - steps;
+            if let Some(proof) = opt {
+                return (nb_calls, proof);
+            }
+        }
+        // If no proof found, return empty proof
```
abailly-iohk, line 295: ditto

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    use super::*;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::{RngCore, SeedableRng};
+
+    #[test]
+    fn test_verify() {
+        let mut rng = ChaCha20Rng::from_seed(Default::default());
+        let nb_tests = 1_000;
+        for _t in 0..nb_tests {
```
abailly-iohk, line 327: Perhaps using a property based testing library would make the test simpler?

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+                r: proof.r,
+                d: proof.d,
+                items: wrong_items.clone(),
+            };
+            assert!(!Proof::verify(&setup, proof_itembis));
+        }
+    }
+
+    #[test]
+    fn test_prove() {
```
abailly-iohk, line 386: Is this really a `test`? Seems like we print stuff at the end too

caledonia/src/prehashed.rs
```diff
@@ -0,0 +1,335 @@
+//! Rust implementation of ALBA's prehashed scheme using Blake2b as hash
+//! function, working for big sets.
+
+extern crate core;
+extern crate utils;
+
+use std::f64::consts::E;
+
+/// Setup input parameters
```
abailly-iohk, line 9: There's a lot of redundancy between this file and the `bounded.rs` one, would be good to DRY it?

caledonia/utils/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "utils"
```
abailly-iohk, line 2: Instead of having another package, a `utils.rs` module in the other package would simplify the whole strucure?

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Approximate size of set Sp to lower bound
+    pub n_p: usize,
+    /// Target lower bound
+    pub n_f: usize,
+}
+/// Setup output parameters
+#[derive(Debug, Clone)]
+pub struct Setup {
+    /// Approximate size of set Sp to lower bound
+    pub n_p: usize,
```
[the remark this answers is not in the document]
rrtoledo: This is indeed redundant with Param. 

We need it in `Param` as it is the input parameter of the setup to compute `u` and the other calues. We also must have `n_p` in `Setup` to pass to the verifier to instantiate the oracle `h1`.

I am not very happy with this solution. Ideally, I would instantiate the oracles in Setup and pass them directly to the prover and verifier. Doing so, I would also not need to pass around hash keys when wanting to personalise the hashes.

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+            b = 1000 * l3;
+            let d_f64 = 16.0 * u_f64 * (lambda_rel + log3) / loge;
+            d = d_f64.ceil() as usize;
+            q = (2.0 * (lambda_rel + log3) / (d_f64 * loge)).recip().ceil() as usize;
+
+            let check = ((d_f64 * d_f64 * loge) / (9.0 * (lambda_rel + log3))).ceil() as usize;
+            assert!(params.n_p >= check);
+        }
+
+        Setup {
```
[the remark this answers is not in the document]
rrtoledo: Just updated when changing the setup with the new case bounds.

caledonia/src/bounded.rs
```diff
@@ -0,0 +1,451 @@
...
+
+/// Round parameters
+#[derive(Debug, Clone)]
+pub struct Round {
+    /// Proof counter
+    v: usize,
+    /// Proof 2nd counter
+    t: usize,
+    // Round candidate tuple
+    s_list: Vec<[u8; 32]>,
```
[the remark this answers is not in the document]
rrtoledo: It would definitely be useful to use types. 

I was thinking of directly using the Hash trait for slist however (https://doc.rust-lang.org/std/hash/trait.Hash.html). 

That way, we wouldn't care about what type of data we are handling as long as it is hashable.

caledonia/src/bounded.rs
rrtoledo: Sure! IIRC, this was a reminder to add the hash of the "s"es to the bins and balls as optimization.

caledonia/src/bounded.rs
rrtoledo: I'll change the name to nb_steps and increment it instead (as well as update the comment).

caledonia/src/bounded.rs
rrtoledo: Perhaps :D Updating it.

caledonia/src/bounded.rs
rrtoledo: Definitely! There is an issue for this. For the time being, it's good to have this test however.

caledonia/src/bounded.rs
rrtoledo: You are correct. This was to get prelim. benches before using criterion (c.f. PR https://github.com/cardano-scaling/alba/pull/15).

caledonia/src/prehashed.rs
abailly-iohk: No, but factor out the common parts in one or more modules?

caledonia/utils/Cargo.toml
rrtoledo: I first had a utils module but changed for some reasons... I will revert it back!

caledonia/src/bounded.rs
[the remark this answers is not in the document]
abailly-iohk: Sound definitely like a good idea!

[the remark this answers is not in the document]
rrtoledo: Unfortunately this trait only gives the choice of a bad hash so I defined types as you first suggested.

caledonia/src/prehashed.rs
rrtoledo: I had a look at it and while the algo is very similar there are too many tiny differences to properly extract something out (like the DFS). Perhaps we could just remove prehashed? 
I'll still need to keep the current structure for the lottery and weight anyway.

abailly-iohk: Sure, no worries

caledonia/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+blake2 = "0.10.6"
+rand_core = "0.6.4"
+
+[dev-dependencies]
+rand = "0.8.5"
+rand_chacha = "0.3.1"
```
djetchev, line 17: newline

caledonia/.gitignore
```diff
@@ -0,0 +1 @@
+target/
```
djetchev, line 1: newline at the end

caledonia/src/prehashed.rs
```diff
@@ -0,0 +1,331 @@
...
+            h: h.to_vec(),
+            h_usize,
+            n_p: r.n_p,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Alba proof
+pub struct Proof {
```
djetchev, line 131: These classes exist in `bounded.rs` too and they share the same interface, so we can use Traits to extract the common functionality

caledonia/src/prehashed.rs
```diff
@@ -0,0 +1,331 @@
...
+
+const DATA_LENGTH: usize = 32;
+const DIGEST_SIZE: usize = 32;
+
+type Data = [u8; DATA_LENGTH];
+type Hash = [u8; DIGEST_SIZE];
+
+/// Setup input parameters
+#[derive(Debug, Clone)]
+pub struct Params {
```
djetchev, line 17: `Params` is common with `bounded.rs` and we could maybe extract the common functionality in a trait?

caledonia/src/prehashed.rs
```diff
@@ -0,0 +1,331 @@
...
+            u,
+            d,
+            q,
+        }
+    }
+}
+
+/// Round parameters
+#[derive(Debug, Clone)]
+pub struct Round {
```
djetchev, line 75: idem

caledonia/src/utils.rs
```diff
@@ -0,0 +1,123 @@
+use blake2::digest::{Update, VariableOutput};
+use blake2::Blake2bVar;
+use std::cmp::min;
+
+// Helper functions
+fn mod_non_power_of_2(hash: &[u8], n: usize) -> usize {
```
djetchev, line 6: add a unit test

abailly-iohk:
I am going to merge this PR as discussed: we need to start somewhere on the path towards a production-ready library and this is as good a starting point as anything.

PR #20, opened 2024-09-18
Make Rust main code in the repository

abailly-iohk:
This inverts the relationship between Rust and Haskell in this repo: The Rust code is the primary code, and Haskell code moves to a subdirectory.


haskell/src/ALBA.hs
djetchev, line 1: @abailly-iohk - is this file going to be deprecated eventually?

caledonia/.gitignore
djetchev, line 1: The `caledonia/` directory gets deprecated.

haskell/src/ALBA.hs
abailly-iohk: Yes

caledonia/.gitignore
abailly-iohk: good catch

abailly-iohk: hmm actually yes, its' been removed

PR #22, opened 2024-09-19
Provide a simple HTML page to compute proof size and time from parameters

ghost:
This PR adds a simple HTML page to the ALBA website that provides user with an interactive model to understand how parameters impact the proof size and adversarial proving time. 

Screenshot 2024-09-19 at 22 42 14


abailly-iohk:
ping @djetchev

abailly-iohk:
As noted by @tolikzinovyev (see https://github.com/cardano-scaling/alba/discussions/17#discussioncomment-10698273) and discussed with @rrtoledo it seems my adversary proof time estimates are way over inflated. My skills in combinatorics being what they are (limited) I would appreciate if someone could suggest a more accurate estimate.

docs/alba.js
```diff
@@ -0,0 +1,136 @@
...
+
+  function updateProofTime() {
+    const n_p_v = Number(n_p.value);
+    const n_f_v = Number(n_f.value);
+    const s_p_v = Number(s_p.value);
+    const cpu_v = Number(cpu.value);
+    const u = U(n_p_v, n_f_v);
+    const proba = probabilityOfProof(u, n_p_v, n_f_v, s_p_v);
+    // expected prover time is n_p + O (u^2), we neglect the n_p part here
+    proof_time.value = (u * u / proba / (cpu_v * HASH_FACTOR)).toExponential(2);
```
amesgen, line 116: Hmm, `proba` is (an upper bound on) the probability that someone with `s_p_v` many elements is able to create a proof. With probability of at least `1 - proba`, they won't be able to create a proof, even if they have infinite computational power. Therefore, I am not sure how one should/could interpret the division by `proba` here :thinking:

tolikzinovyev:
> As noted by @tolikzinovyev (see [#17 (reply in thread)](https://github.com/cardano-scaling/alba/discussions/17#discussioncomment-10698273)) and discussed with @rrtoledo it seems my adversary proof time estimates are way over inflated. My skills in combinatorics being what they are (limited) I would appreciate if someone could suggest a more accurate estimate.

Basically, there are two scenarios. Either there exists an adversarial proof and the adversary can find it very quickly, or there doesn't exists an adversarial proof and so the adversary cannot find one no matter how much computation is done. So I think modeling CPU power is not useful for ALBA.

docs/alba.js
abailly-iohk: yeah, I am mixing apples and oranges. I will just show the probability for a single $S_p$  value and that should be it.

abailly-iohk:
So it's just a matter of luck, and there's no way to brute force finding a proof?

tolikzinovyev:
Only if grinding proof statement is possible. E.g., the attacker can move his money to new accounts thus creating new signatures. In that case we need to rehash the blockchain state and include it as a seed to the hash function used by ALBA. Then the security becomes computation as opposed to information theoretic. But it kind of is outside the scope for ALBA.

docs/src/intro.md
```diff
@@ -1,44 +1,17 @@
 # Introduction
 
-_Approximate Lower Bound Arguments_ are a form of cryptographic certificates that allow a _prover_ to convince a _verifier_ they know some large set of elements by providing only a small subset of those elements. This excellent [X thread](https://x.com/Quantumplation/status/1783188333046255997) provides a good intuition on why ALBAs are useful and how they work, and details of the theory behind this construction are beyond the scope of this introduction and can be found in the paper.
+## Definition
 
-### Proving
+_Approximate Lower Bound Arguments_ are a form of cryptographic certificates that allow a _prover_ to convince a _verifier_ they know some set of elements by providing only a small subset of those elements. More formally, given some set of _unique_ elements, and some bounds \\(n_p\\) and \\(n_f\\), where \\(n_p > n_f\\), about the size of this set, ALBA provides an algorithm that allows to build a _proof_ they know _at least_ \\(n_f\\) elements, where the proof size is a constant value derived from the \\(\frac{n_p}{n_f}\\) ratio.
```
amesgen, line 5: > given some set of _unique_ elements

What do you mean by this? What would be an example of a set of *non-unique* elements?

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
+# ALBA Parameters Simulation
+
+ALBA provides strong security guarantees about the capability to generate a proof given some number of items \\(S_p\\):
+
+* a honest prover with more than \\(n_p\\) items has a probability of
```
amesgen, line 5:
```suggestion
* a honest prover with at least \\(n_p\\) items has a probability of
```

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
...
+  failing to generate a proof lower than \\(2^{-128}\\),
+* symetrically, a (dishonest) prover with less than \\(n_f\\) items
```
amesgen, line 7:
```suggestion
* symetrically, a (dishonest) prover with at most \\(n_f\\) items
```

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
...
+
+From these parameters, we display a chart showing the probability of
+an adversary being able to compute a single valid proof, given some
+number of items \\(S_p\\) available between \\(n_p\\) and
+\\(n_f\\). The formula used, initially proposed in this [GitHub
+discussion](https://github.com/cardano-scaling/alba/discussions/17)
+is:
+
+\\[
+ r \cdot {\left( \frac{|S_p|}{n_p} \right)}^u\cdot qd
```
amesgen, line 34:
```suggestion
 r \cdot {\left( \frac{S_p}{n_p} \right)}^u\cdot qd
```
given your definition of $S_p$ as a number instead of a set above

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
...
+
+One can define:
+
+* \\(n_p\\): The size of the _honest set_ of items.
+* \\(n_f\\): The size of the _faulty set_ of items. \\(n_f\\) must be lower than \\(n_p\\).
+* Item size: The size of a single item, this is used to compute the full size of a proof.
+* \\(S_p\\): The number of items actually available to an adversary, or put differently, how much adversarial stake should we expect.
+
+From these parameters, we display a chart showing the probability of
+an adversary being able to compute a single valid proof, given some
```
amesgen, line 27:
```suggestion
an adversary being able to compute a valid proof, given some
```
The probability term bounds the probability that an adversary can create *at least one* valid proof.

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
...
+where _u_, _r_, _q_, and _d_ are defined in the aforementioned
+section of the paper. Note the vertical axis' scale is logarithmic.
+
+We also display two specific values:
+
+* the size of the proof for the given \\(n_p/n_f\\) ratio and individual
+  item size, which is given by the formula from section 3.2.2,
+  Corollary 3, from the [ALBA
+  paper](https://iohk.io/en/research/library/papers/approximate-lower-bound-arguments/),
+* the probability of having a proof for some specific number of \\(S_p\\) (computed.
```
amesgen, line 46:
```suggestion
* the probability of having a proof for some specific number of \\(S_p\\) computed.
```

docs/src/simulation.md
```diff
@@ -0,0 +1,84 @@
...
+estimate what would be "good" values for those parameters in a
+specific context, depending on security requirements and expected
+adversarial share.
+
+One can define:
+
+* \\(n_p\\): The size of the _honest set_ of items.
+* \\(n_f\\): The size of the _faulty set_ of items. \\(n_f\\) must be lower than \\(n_p\\).
+* Item size: The size of a single item, this is used to compute the full size of a proof.
+* \\(S_p\\): The number of items actually available to an adversary, or put differently, how much adversarial stake should we expect.
```
amesgen, line 24: > $S_p$: The number of items actually available to an adversary, or put differently, how much adversarial stake should we expect.

Hmm, depending on the context and parameters, the adversary might still be able to do weird things even if they created none of the $S_p$ many items; for example, in the context of Peras, citing from https://github.com/cardano-scaling/alba/discussions/17:

> Suppose that honest participants only attempt to create an ALBA proof when they have at least some minimum number of votes. Suppose that an adversary is eligible for sufficiently many votes such that the honest participants do not reach this threshold. Then the adversary might still be able to create a valid ALBA proof, causing confusion at the very least.

Phrased differently: An adversary might be able to create a certificate signalling a quorum despite the quorum having been missed, using only honest votes.

docs/alba.js
```diff
@@ -0,0 +1,133 @@
...
+
+  function probabilityOfProof(u, n_p, n_f, x) {
+    // from https://github.com/cardano-scaling/alba/discussions/17
+
+    S_1_bot = n_p / (17 ** 2 / (9 * Math.log2(Math.E)) * u ** 2) - 7 < 1;
+    S_2_bot = n_p / (17 ** 2 / (9 * Math.log2(Math.E)) * u ** 2) - 2 < 1;
+
+    if (!S_1_bot && !S_2_bot) {
+      throw Error("S_1_bot or S_2_bot");
+    }
```
amesgen, line 26: - Currently, this will just silently log this error to the console, potentially confusing users. Maybe use `alert` or sth like that?
 - One could also implement the remaining two cases in the proof of Corollary 3 in order to handle large `n_p` values (where this error is currently thrown).

(Doesn't necessarily have to happen in this PR)

docs/alba.js
```diff
@@ -0,0 +1,133 @@
+document.addEventListener('DOMContentLoaded', () => {
+
+  const lam = 128;
+  const HASH_FACTOR = 1e8;
```
amesgen, line 4: Now unused
```suggestion
```

docs/alba.js
```diff
@@ -0,0 +1,133 @@
...
+
+    if (!S_1_bot && !S_2_bot) {
+      throw Error("S_1_bot or S_2_bot");
+    }
+
+    const d = Math.ceil(32 * Math.log(12) * u);
+    const q = 2 * Math.log(12) / d;
+    const r = Math.ceil(lam);
+
+    return r * (x / n_p) ** u * d * q;
```
amesgen, line 32:
```suggestion
    return Math.min(r * (x / n_p) ** u * d * q, 1);
```
This (expectedly, due to the union bound) returns values larger than 100% for large `n_p`. Alternatively, the plot can be truncated at 100% on the y axis (as in https://github.com/cardano-scaling/alba/discussions/17).

docs/alba.js
```diff
@@ -0,0 +1,133 @@
...
+    proof_size.value = u * single;
+  }
+
+  function updateProofProba() {
+    const n_p_v = Number(n_p.value);
+    const n_f_v = Number(n_f.value);
+    const s_p_v = Number(s_p.value);
+    const u = U(n_p_v, n_f_v);
+    const proba = probabilityOfProof(u, n_p_v, n_f_v, s_p_v);
+    // expected prover time is n_p + O (u^2), we neglect the n_p part here
```
amesgen, line 113:
```suggestion
```
(superseded)

docs/src/intro.md
abailly-iohk: Silly me 🤦

abailly-iohk: this is my dumb rephrasing of the ALBA requirement that $S_p$ must be indeed a set, so that one cannot grow arbitrarily the size of items to construct a proof from. This would be a problem in the case of simple Ed25519 signatures because, as I understand from discussions with cryptographers, anyone can create more valid signatures from a single signature. this is not a problem in general, but for ALBA it obviously is.

docs/alba.js
abailly-iohk: Yes, I copy/pasted your code. I think it's good enough for now but obviously throwing exceptions in user-facing code is not great. I don't like `alert()` but can find a good substitute.

abailly-iohk:
Thanks for the thorough review @amesgen, I will fix the remaining issues tomorrow.

docs/src/intro.md
amesgen: Ah, makes sense, so this paragraph from section 1.3 from the paper:



That indeed seems worth to point out somewhere in more detail :+1:

docs/src/simulation.md
abailly-iohk: I will simplify the sentence, because what $S_p$ is in this context is just the number of items one has at their disposal to create a certificate and there's not really any notion of an adversary anymore.

docs/alba.js
rrtoledo: Why do you throw an error and not update the d/q/r parameters depending on S1 and S2?

tolikzinovyev:
The simulation looks nice. I would just use a different name for S_p. S_p is normally defined as the honest party's elements. Also, S_p is a set, not a number. Maybe we just call it n?

docs/alba.js
tolikzinovyev: I think for this simulation we can just use the parameters for the "small case". They in fact work for all cases, it's just, the prover will not have a good worst-case running time. Then we don't need to worry about S_1_bot, etc.

We can borrow the math code from Raphael's implementation.
https://github.com/cardano-scaling/alba/pull/33/files#diff-b0bad697f839908964cc4e445cd9c87a76b4a5f84da347b24a426b61b7d09322R127

docs/alba.js
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  function updateProofProba() {
+    const n_p_v = Number(n_p.value);
+    const n_f_v = Number(n_f.value);
+    const n_v = Number(n.value);
+    const u = U(n_p_v, n_f_v);
+    const proba = probabilityOfProof(u, n_p_v, n_f_v, n_v);
+    // expected prover time is n_p + O (u^2), we neglect the n_p part here
+    proof_proba.value = proba.toExponential(4);
```
tolikzinovyev, line 113: Confused about what this does.

rrtoledo:
Let's merge this with @curiecrypt ongoing documentation efforts

PR #31, opened 2024-10-08
Raph@generics

rrtoledo:
## Content

This PR includes...

## Pre-submit checklist

- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
- PR
    - [ ] No clippy warnings in the CI
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reviewer requested
- Documentation

## Issue(s)

Relates to #YYY or Closes #YYY

### DIFF
--- a/src/centralized_telescope/algorithm.rs
+++ b/src/centralized_telescope/algorithm.rs
@@ -12,7 +12,7 @@ use blake2::{Blake2s256, Digest};
 /// proof if no suitable candidate is found.
 pub fn prove<Element>(setup: &Setup, prover_set: &[Element]) -> Option<Proof<Element>>
 where
-    Element: Copy + Clone + ToBytes,
+    Element: Copy + ToBytes,
 {
     // Run prove_index up to max_retries times
     (0..setup.max_retries).find_map(|retry_counter| prove_index(setup, prover_set, retry_counter).1)
@@ -22,7 +22,7 @@ where
 /// successfully verified, following the DFS verification, false otherwise.
 pub fn verify<Element>(setup: &Setup, proof: &Proof<Element>) -> bool
 where
-    Element: Copy + Clone + ToBytes,
+    Element: Copy + ToBytes,
 {
     if proof.search_counter >= setup.search_width
         || proof.retry_counter >= setup.max_retries
@@ -41,7 +41,7 @@ where
     // For each element in the proof's sequence
     for &element in &proof.element_sequence {
         // Retrieve the bin id associated to this new element
-        let Some(bin_id) = bin_hash(setup, proof.retry_counter, element) else {
+        let Some(bin_id) = bin_hash(setup, proof.retry_counter, &element) else {
             return false;
         };
         // Check that the new element was chosen correctly
@@ -67,7 +67,7 @@ fn prove_index<Element>(
     retry_counter: u64,
 ) -> (u64, Option<Proof<Element>>)
 where
-    Element: Copy + Clone + ToBytes,
+    Element: Copy + ToBytes,
 {
     // Initialise set_size bins
     let mut bins: Vec<Vec<Element>> = Vec::with_capacity(setup.set_size as usize);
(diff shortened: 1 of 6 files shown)


caledonia/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
```
djetchev, line 1: This is a basic question, but should we remove the `caledonia/` folder as it's no longer needed: project-root/
```
├── benches/
├── examples/ 
├── src/
```

rrtoledo:
@tolikzinovyev @curiecrypt @djetchev
I would put back the functions in an `impl Proof` to simplify all the `fn name<Element>() where Element ...`.
Should I keep "<Element>" or use the more common "<T>"?

tolikzinovyev:
@rrtoledo there are probably many ways to implement generic data elements. Could you say what other options you considered and why you chose this one? Thanks.

src/centralized_telescope/algorithm.rs
```diff
@@ -4,21 +4,26 @@ use super::proof::Proof;
...
-use crate::utils::sample;
-use crate::utils::types::Element;
+use crate::utils::{sample, types::ToBytes};
 use blake2::{Blake2s256, Digest};
 
 /// Alba's proving algorithm, based on a depth-first search algorithm.
 /// Calls up to setup.max_retries times the prove_index function and returns an empty
 /// proof if no suitable candidate is found.
-pub fn prove(setup: &Setup, prover_set: &[Element]) -> Option<Proof> {
+pub fn prove<Element>(setup: &Setup, prover_set: &[Element]) -> Option<Proof<Element>>
```
djetchev, line 13: @rrtoledo @tolikzinovyev @curiecrypt : if I think about it, the most general setting occurs when `prove()` is parameterized by a generic type and this generic type under the constraint that this type implemetns the traits Eq + Hash. Can we do something like this 
```
fn prove<T: Eq + Hash>()
```

djetchev: If you don't want to write `<T: Eq + Hash>` in every function signature, you can first define 
```
use std::hash::Hash;
use std::collections::HashSet;

trait HashableEq: Eq + Hash {}

impl<T: Eq + Hash> HashableEq for T {}
```
This implements `HashableEq` for all types `T` that requires implementing both `Eq` and `Hash` and the second line ensures that any type that implements Eq and Hash automatically satisfies `HashableEq`, so you can now write 
```
fn prove<T: HashableEq>(...) {}
```

src/centralized_telescope/algorithm.rs
```diff
@@ -4,21 +4,26 @@ use super::proof::Proof;
...
+    Element: Copy + Clone + ToBytes,
+{
     // Run prove_index up to max_retries times
     (0..setup.max_retries).find_map(|retry_counter| prove_index(setup, prover_set, retry_counter).1)
 }
 
 /// Alba's verification algorithm, returns true if the proof is
 /// successfully verified, following the DFS verification, false otherwise.
-pub fn verify(setup: &Setup, proof: &Proof) -> bool {
+pub fn verify<Element>(setup: &Setup, proof: &Proof<Element>) -> bool
```
djetchev, line 23: idem

src/centralized_telescope/algorithm.rs
rrtoledo: @djetchev This looks interesting, we would also need the Copy trait however.

I had a look at the `Hash` trait in the past, and I wanted to stay clear from it as it does not seem it was meant for cryptographic purposes (but DoS for hashmaps).

If you look at [how it works](https://doc.rust-lang.org/std/hash/trait.Hash.html), you need to use a `Hasher` in conjunction to it. However the [default hasher](https://github.com/rust-lang/rust/blob/master/library/std/src/hash/random.rs#L110) is SipHasher13 which is not cryptographically secure but used only to prevent DOS in hashmaps.

I am afraid that by using this trait, we are encourage the use of this default hasher. Perhaps I am being overlycautious, and we can simply provide a different default hash.

rrtoledo:
Closing this PR temporarily to start others of higher priority

PR #33, opened 2024-10-08
Adding centralised alba (bounded scheme)

rrtoledo:
## Content
Adding Bounded scheme (section 3.2.2 ) implementation in Rust with randomized tests and benches.

This PR includes...

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
- Documentation

## Comments
Repetition bench doesn't work so is disabled

## Issue(s)
Issue #25


benches/centralised_repetition.rs
```diff
@@ -0,0 +1,85 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use rand_chacha::ChaCha20Rng;
+use rand_core::SeedableRng;
+use std::time::Duration;
+
+use caledonia::centralised::Proof;
```
djetchev, line 6: remove reference to `caledonia`

benches/centralised_repetition.rs
```diff
@@ -0,0 +1,85 @@
...
+pub mod utils;
+
+fn prepetitions(
+    c: &mut Criterion<criterion_helpers::Repetitions>,
+    lambdas: &[usize],
+    s_p: &[usize],
+    n_p: &[usize],
+    n_f: &[usize],
+) {
+    let mut group = c.benchmark_group("Centralised".to_string());
```
djetchev, line 18: Centralised --> Centralized (and identically in the file name too and other places)

benches/centralised_repetition.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    fn prove_repetitions(
```
djetchev, line 20: Would be nice to add some high level comment on what this function is doing exactly.

benches/centralised_repetition.rs
```diff
@@ -0,0 +1,85 @@
...
+use rand_chacha::ChaCha20Rng;
+use rand_core::SeedableRng;
+use std::time::Duration;
+
+use caledonia::centralised::Proof;
+
+pub mod criterion_helpers;
+pub mod utils;
+
+fn prepetitions(
```
djetchev, line 11: Explain what this function is doing.

src/centralised.rs
```diff
@@ -1,12 +1,16 @@
-//! Rust implementation of ALBA's bounded DFS scheme using Blake2b as hash
-//! function.
+//! ALBA's bounded DFS scheme using Blake2b as hash function.
```
djetchev, line 1: change name to `centralized.rs`

src/utils.rs
```diff
@@ -80,6 +83,67 @@ pub fn gen_items<const N: usize>(seed: Vec<u8>, set_size: usize) -> Vec<[u8; N]>
     s_p
 }
 
+/// Generate a set of weighted items given the total weight size, number of items and a seed
```
djetchev, line 86: What is an item in this context? I'd like to better understand the purpose of this function and how it is used. You seem to be hashing the seed + the current index.

benches/utils.rs
```diff
@@ -0,0 +1,24 @@
...
+use caledonia::utils::gen_items;
+
+pub fn setup_centralised_wrapper(
+    rng: &mut ChaCha20Rng,
+    l: usize,
+    sp: usize,
+    np: usize,
+    nf: usize,
+) -> (Vec<[u8; 32]>, caledonia::centralised::Setup) {
+    use caledonia::centralised::*;
```
djetchev, line 13: reference to Caledonia?

benches/utils.rs
```diff
@@ -0,0 +1,24 @@
+use rand_chacha::ChaCha20Rng;
+use rand_core::RngCore;
+
+use caledonia::utils::gen_items;
```
djetchev, line 4: ref to Caledonia

curiecrypt (commented):
First round of the review. I will do one more time for benches.

benches/centralised_repetition.rs
rrtoledo: Sorry, I forgot to write in American English!

src/utils.rs
rrtoledo: This function is under the "// test & Bench helpers" comment, is that not enough?

I'll create a test folder sure.

p.s. The weighted function will be removed, that was an oversight.

src/utils.rs
```diff
@@ -80,6 +83,67 @@ pub fn gen_items<const N: usize>(seed: Vec<u8>, set_size: usize) -> Vec<[u8; N]>
...
+/// Items are generated by hashing the current index
+pub fn gen_weighted_items<const N: usize>(
```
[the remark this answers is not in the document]
rrtoledo: This is an oversight and has just been removed in a commit

src/centralised.rs
```diff
@@ -318,19 +322,19 @@ impl Proof {
         for &s in set.iter() {
             bins[Proof::h0(setup, v, s)].push(s);
         }
-        let mut nb_steps = 0;
+        let nb_steps = Arc::new(AtomicUsize::new(0));
```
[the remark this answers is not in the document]
rrtoledo: I changed the name from limit as I am incrementing the variable instead of decreasing it. If you think it's still fine, I'll rename it to limit.

benches/centralised_repetition.rs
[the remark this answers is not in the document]
rrtoledo: It was quite hard to do so (as we do benches on different measurements) but I managed to do it and pushed something.

tolikzinovyev:
I cannot comment on the lines that weren't changed (the gray lines). :(   I really think we should remove all the code from the main branch and then review. Otherwise, reviewing is much harder.

tolikzinovyev:
This is a nice first implementation, Raphael. I've looked over `centralized.rs` and I wonder if we should modify the approach a bit. If this library intends to eventually become production ready with high guarantees of security (and I really hope it does), it might make sense to make the first PR a very small simple implementation to make sure everyone can review it and ascertain correctness with ease. We should prioritize code readability and good interfaces and leave optimizations such as parallelization and reusing H_1 hashes for future, also short, PRs. In other words, making small steps that are easy to verify.

I admit that I'm generally very pedantic about code, but given that it's a cryptographic library, it makes sense to stay on the safe side. This doesn't mean, however, that we can't move fast, and I'm also happy to volunteer a tiny first implementation, if you allow me.

rrtoledo:
> I cannot comment on the lines that weren't changed (the gray lines). :( I really think we should remove all the code from the main branch and then review. Otherwise, reviewing is much harder.

@tolikzinovyev all of the old rust code has been removed in this PR. If you are experiencing difficulties reviewing the commits  that have been added, it is part of the reviewing process to say why so that we can go through it together and improve in the future.

As for the untouched code, you can still select lines and comment on them when looking at the files though the "commits" or "file changed" tabs here. Were you not to be able to do so, you can still go through the code and write your points in a general comment here or on a paper and we would then be able to look at them together.

tolikzinovyev:
> > I cannot comment on the lines that weren't changed (the gray lines). :( I really think we should remove all the code from the main branch and then review. Otherwise, reviewing is much harder.
> 
> @tolikzinovyev all of the old rust code has been removed in this PR. If you are experiencing difficulties reviewing the commits that have been added, it is part of the reviewing process to say why so that we can go through it together and improve in the future.
> 
> As for the untouched code, you can still select lines and comment on them when looking at the files though the "commits" or "file changed" tabs here. Were you not to be able to do so, you can still go through the code and write your points in a general comment here or on a paper and we would then be able to look at them together.

Never mind, I just checked that I can leave comments on unchanged lines. It's weird that I couldn't before.

rrtoledo:
> This is a nice first implementation, Raphael. I've looked over `centralized.rs` and I wonder if we should modify the approach a bit. If this library intends to eventually become production ready with high guarantees of security (and I really hope it does), it might make sense to make the first PR a very small simple implementation to make sure everyone can review it and ascertain correctness with ease. We should prioritize code readability and good interfaces and leave optimizations such as parallelization and reusing H_1 hashes for future, also short, PRs. In other words, making small steps that are easy to verify.
> 
> I admit that I'm generally very pedantic about code, but given that it's a cryptographic library, it makes sense to stay on the safe side. This doesn't mean, however, that we can't move fast, and I'm also happy to volunteer a tiny first implementation, if you allow me.

@tolikzinovyev thanks for looking at the implementation of the centralised scheme.

We will go through many iterations of the code indeed and having small PRs is easier to manage these, I agree, but we have to start from somewhere and do not want to re-invent the wheel. We have code, we should reuse it, review it and update it.
Alba centralised scheme is not a complexe nor big, we want here to assert that these 400 lines of code follow the paper's pseudocode, and check that the tests and benchmarks make sense. This should be quite straightforward. Further testing, including property testing, and optimization are taking into account in our roadmap and shall be dealt with other issues and PRs.

Please write down your issues on specific lines so that we can discuss each point separately. As for parallelization, I will remove it as indeed it is a bit early to have it here. There was anyway another PR schedule to deal with the limitation of the current one and I would be more than happy to work with you then on it as we already discussed on how to imrpove it. I am not sure what you are talking about "reusing H1", please give more details on the corresponding lines.

This is a cryptographic library and I agree we have to be careful about what we publish. We are however using versioning, and are not currently releasing version 1.0 of Alba. We can hence afford to go step by step and, to begin with, merge code that corresponds to the paper's pseudocode and again improve it later on. We work by proposing code, reviewing it and refining it.

We need to move forward, and cannot afford to have everyone work on their own implementation, however small it may be. Design choices need to be made, they can and should of course be discussed, but when working in a team we may not all agree and need to compromise. This is what the reviewing process is for.
I encourage you to write here how you would have done things differently, why and what benefits these changes would bring us. We can also go though the code together and have this discussion in a call as these tend to be more efficient.

tolikzinovyev:
By reusing h1, I mean the way h1 hashes are computed. In the paper h1 has long inputs: h1(v, t, s1, ..., s_k), but you update h1 recursively: h2(h1(h1(v, t), s1), s2). This is the reason why your implementation doesn't correspond the paper's specification. This is indeed an important optimization, I'm just suggesting we do it in a separate PR after we are confident about the most basic implementation.

By the way, we never wrote a proof that this optimization works and there is no specification. This is a todo item on the research side.

tolikzinovyev:
Found those gray lines. I cannot leave comments on lines 17-24.

Screenshot 2024-10-17 at 09-05-54 Adding centralised alba (bounded scheme) by rrtoledo · Pull Request #33 · cardano-scaling_alba

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::{RngCore, SeedableRng};
+
+    #[test]
+    fn test_params() {
```
tolikzinovyev, line 393: This function doesn't seem to test anything -- there are no `assert`s. Why do we have it?

src/utils.rs
```diff
@@ -0,0 +1,114 @@
+use blake2::digest::{Update, VariableOutput};
+use blake2::Blake2bVar;
+use std::cmp::min;
+
+// Oracles
+
+/// Takes as input a hash and inverse probability $np$ and returns a
```
tolikzinovyev, line 7: inverse probability?

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+/// uniformly distributed integer in [1, np] (c.f. Appendix B, Alba paper).
+pub fn oracle_uniform(hash: &[u8], n: usize) -> usize {
+    // Computes the integer reprensation of hash* modulo n when n is not a
+    // power of two. *(up to 8 bytes, in little endian)
+    fn mod_non_power_of_2(hash: &[u8], n: usize) -> usize {
+        fn log_base2(x: usize) -> usize {
+            usize::BITS as usize - x.leading_zeros() as usize - 1
+        }
+        let epsilon_fail: usize = 1 << 40; // roughly 1 trillion
```
tolikzinovyev, line 16: Similarly, we can't use `usize` here since on 32-bit architectures it will not fit 2^40.

tolikzinovyev, line 16: One trillion seems low compared to typical security / reliability parameters. Is this going to be improved in the future?

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+        let k: usize = log_base2(n * epsilon_fail);
+        let k_prime: usize = 1 << k;
+        let d: usize = k_prime.div_ceil(n);
+
+        let i = mod_power_of_2(hash, k_prime);
+
+        if i >= d * n {
+            panic!("failed: i = {}, d = {}, n = {}, k = {}", i, d, n, k);
```
tolikzinovyev, line 24: I think we should return an `Option` from this function instead of panicking, similarly to the prove() function.

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+        mod_power_of_2(hash, n)
+    } else {
+        mod_non_power_of_2(hash, n)
+    }
+}
+
+/// Takes as input a hash and probability $q$ and returns true with
+/// probability q otherwise false according to a Binomial distribution
+/// (c.f. Appendix B, Alba paper).
+pub fn oracle_binomial(hash: &[u8], q: f64) -> bool {
```
tolikzinovyev, line 46: binomial -> bernoulli

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+    let bytes = &bytes[..min(8, bytes.len())];
+    array[..bytes.len()].copy_from_slice(bytes);
+    usize::from_le_bytes(array)
+}
+
+// Hash helpers
+
+/// Return a N-byte long hash of the given data
+pub fn hash_bytes<const N: usize>(data: &[u8]) -> [u8; N] {
+    let mut hasher = Blake2bVar::new(N).expect("Failed to construct hasher!");
```
tolikzinovyev, line 79: When can we in principal get an error? I would like to understand the risk of crashing here.

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+        .finalize_variable(&mut buf)
+        .expect("Failed to finalize hashing");
+    buf
+}
+
+// Test & Bench helpers
+
+/// Generate a set of items given the set size and a seed
+/// Items are generated by hashing the current index
+pub fn gen_items<const N: usize>(seed: Vec<u8>, set_size: usize) -> Vec<[u8; N]> {
```
tolikzinovyev, line 105: Suggest moving it to `test_utils.rs` or a similar file.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            n_p: r.n_p,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Alba proof
+pub struct Proof {
+    /// Proof counter
+    r: usize,
```
tolikzinovyev, line 245: We should call the variables `v`, `t`, `elements` to be consistent with the paper.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
+//! ALBA's bounded DFS scheme using Blake2b as hash function.
```
tolikzinovyev, line 1: Suggest renaming the file to `centralized_telescope.rs` since we will have the simple lottery scheme as well.

src/utils.rs
```diff
@@ -0,0 +1,114 @@
+use blake2::digest::{Update, VariableOutput};
+use blake2::Blake2bVar;
+use std::cmp::min;
+
+// Oracles
+
+/// Takes as input a hash and inverse probability $np$ and returns a
+/// uniformly distributed integer in [1, np] (c.f. Appendix B, Alba paper).
+pub fn oracle_uniform(hash: &[u8], n: usize) -> usize {
```
tolikzinovyev, line 9: "oracle" doesn't seem accurate since this function doesn't produce randomness. I would maybe call it "sample"

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+pub fn gen_items<const N: usize>(seed: Vec<u8>, set_size: usize) -> Vec<[u8; N]> {
+    let mut s_p = Vec::with_capacity(set_size);
+    for b in 0..set_size {
+        let mut data = vec![seed.clone()];
+        data.push(b.to_ne_bytes().to_vec());
+        let item = combine_hashes::<N>(data);
+        s_p.push(item);
+    }
+    s_p
+}
```
tolikzinovyev, line 114: nit: it would be nice to have a consistent function order, either call tree or reverse call tree. https://www.staycaffeinated.com/2021/05/15/coding-standards-function-ordering

tolikzinovyev (commented):
First set of comments. I'm going to review the math in parameter derivation next.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Computation bound
+    pub b: usize,
+}
+impl Setup {
+    /// Setup algorithm taking a Params as input and returning setup parameters (u,d,q)
+    pub fn new(params: &Params) -> Self {
+        let loge = E.log2();
+        fn compute_w(u: f64, l: f64) -> f64 {
+            fn factorial_check(w: f64, l: f64) -> bool {
+                let bound = 0.5f64.powf(l);
```
tolikzinovyev, line 92: suggest using specialized function: `(-l).exp2()`

tolikzinovyev (commented):
Finished reviewing the math in parameter derivation expect for `factorial_check()`. Things mostly look good, left some comments.

src/centralized.rs
djetchev: @tolikzinovyev , @rrtoledo : How about we give them a more descriptive namings in the code and explain in comments how they map to the notation in the paper?

src/utils.rs
djetchev: Also, with the risk of being pedantic :), I'd like to see a slightly more clear comment - we don't need to say that n_p is an inverse probability, but we should maybe state how we use the hash.

src/utils.rs
djetchev: Let's rename the function appropriately and fix the comment appropriately. Also, I'd briefly explain how the hash is used.

src/utils.rs
djetchev: `Option` or `Result<T, E>`?

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        if i >= d * n {
+            panic!("failed: i = {}, d = {}, n = {}, k = {}", i, d, n, k);
+        } else {
+            i % n
+        }
+    }
+    // Computes the integer reprensation of hash* modulo n when n is a power of
+    // two. *(up to 8 bytes, in little endian)
+    fn mod_power_of_2(hash: &[u8], n: usize) -> usize {
```
djetchev, line 31: what is the reason for calling the function parameter `hash`? Isn't that just an array of bytes?

src/utils.rs
tolikzinovyev: Just `Option` I think. There is only one reason to fail, and this error will not propagated to the user anyway since the DFS needs to continue.

tolikzinovyev: We could also reference section B in the paper that talks about it.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+                let factors: Vec<u64> = (1..=(w as u64 + 2)).rev().collect();
```
tolikzinovyev, line 93: There is no reason to collect elements in a vector
```suggestion
                let factors = (1..=(w as u64 + 2)).rev();
```

tolikzinovyev, line 93: We need to divide by (w+1)! instead of (w+2)! based on the paper.

Cargo.toml
```diff
@@ -1,9 +1,29 @@
...
+
+[[bench]]
+name = "centralized_time"
+harness = false
+path = "benches/centralized.rs"
+
+[[bench]]
+name = "centralized_step"
+harness = false
+path = "benches/centralized.rs"
```
djetchev, line 29: add newline at the end of the file

src/lib.rs
```diff
@@ -0,0 +1,16 @@
+//! Approximate Lower Bound Arguments (ALBA, <https://eprint.iacr.org/2023/1655.pdf>)
+//!
+//! Alba is a generic protocol to prove succinctly a lower bound of the size of
+//! a, potentially weighted, set. Say we have a set Sp of size |Sp| >= $n_p$,
```
djetchev, line 4: by a `potentially weighted` set do you simply mean `multiset`?

src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+//! and a lower bound $n_f$ < $n_p$ of it we want to prove. Alba gives us a
+//! method to generate a proof of knowledge of this bound by finding the
```
djetchev, line 6: `the smallest` - is it unique? Is it really smaller in the sense that it cannot contain a smaller set that also convinces a verifier ?

src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+//! smallest subset of Sp of size $u$ to convince a verifier.
+//! The paper presents several schemes and optimizations. The basic scheme is
+//! enhanced in the "prehashed" version thanks to sorting Sp with a balls and
```
djetchev, line 9: The reason why we call this `prehashed` is that we precompute the hashes of the bins in advance. Maybe we could make this clear.

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+    hasher
+        .finalize_variable(&mut buf)
+        .expect("Failed to finalize hashing");
+    buf
+}
+
+/// Return a N-byte long hash of the given list of data
+pub fn combine_hashes<const N: usize>(hash_list: Vec<Vec<u8>>) -> [u8; N] {
+    let mut hasher = Blake2bVar::new(N).expect("Failed to construct hasher!");
+    for data in hash_list.iter() {
```
djetchev, line 91: do you need to call `.iter()` here ?

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+        hasher.update(data);
+    }
+    let mut buf = [0u8; N];
+    hasher
+        .finalize_variable(&mut buf)
```
djetchev, line 96: similarly, if you allow the caller to handle errors, just say 
```
hasher.finalize_variable(&mut buf)?
```

src/utils.rs
```diff
@@ -0,0 +1,114 @@
...
+        .expect("Failed to finalize hashing");
+    buf
```
djetchev, line 98: If caller error handling, replace by 
```
Ok(buf)
```

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+        utils::oracle_binomial(&digest, setup.q)
+    }
+
+    /// Depth-first search which goes through all potential round candidates
+    /// and returns first round candidate Round{t, x_1, ..., x_u)} such that:
+    /// - for all i ∈ [0, u-1], H0(x_i+1) ∈ bins[H1(t, x_1, ..., x_i)]
+    /// - H2(t, x_0, ..., x_u) = true
+    fn dfs(
+        setup: &Setup,
+        bins: &Vec<Vec<Data>>,
```
djetchev, line 288: In Rust, it is preferable to use references to a slice (`&[Vec<Data>]`) rather than `&Vec<Vec<Data>>` unless you are explicitly interacting with some of the features of `Vec` (i.e., pushing or resizing). This is slightly more performant since you are not tying your parameter via the vector (allocated on the heap) but only to a contiguous portion of it. You are therefore avoiding some of the unnecessary heap management related to handling a `Vec`.  This gives you more flexibility too since you can pass a slice of an array to the function.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            if Proof::h2(setup, round) {
+                let r = round.v;
+                let d = round.t;
+                let items = round.s_list.clone();
+                return Some(Proof { r, d, items });
+            } else {
+                return None;
+            }
+        }
+        let result = bins[round.h_usize].par_iter().find_map_first(|&s| {
```
djetchev, line 302: You can simplify the return statement by simply replacing lines 301--309 by 
```
    bins[round.h_usize]
        .par_iter()
        .find_map_first(|&s| {
            // Stop if nb_steps has reached the threshold
            if nb_steps.load(atomic::Ordering::Relaxed) == setup.d {
                return None;
            }

            // Increment the step counter
            nb_steps.fetch_add(1, atomic::Ordering::Relaxed);

            // Recursive DFS with updated round
            Self::dfs(setup, bins, &Round::update(round, s), Arc::clone(&nb_steps))
        })
```

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Proof max 2nd counter
+    pub d: usize,
+    /// Probability q
+    pub q: f64,
+    /// Computation bound
+    pub b: usize,
+}
+impl Setup {
+    /// Setup algorithm taking a Params as input and returning setup parameters (u,d,q)
+    pub fn new(params: &Params) -> Self {
```
djetchev, line 88: I'd think about ways to refactor this constructor to make the code a bit more readable and also, avoid code duplications. Here are a few ideas: 

1) extract a function to compute the u parameter outside of the constructor and simply call this function 

2) extract a function (e.g., `foo()`)  to determine which case we are into (Theorem 13, 14, etc.). This function can give you back also the appropriate parameters for each case. 

3) Use a match expression to construct the different setup parameters. Something like 

```
match foo() {
  Case::Small { <params for small> } {
     // construct the Setup object for the small case 
  }
  // other cases
}
```

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            (Cases::Mid, u as usize)
+        } else {
+            (Cases::High, u as usize)
+        }
+    }
+}
+
+/// Setup output parameters
+#[derive(Debug, Clone)]
+pub struct Setup {
```
djetchev, line 72: In this structure, I'd definitely recommend using more descriptive names that are readable by someone from the outside. 

`set_size` <> `n_p`  
`proof_size` <> `u`

the counters should be better explained too. 

`num_dfs_samples` <> `d` 

etc. etc.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+                    .floor() as usize,
+            }
+        }
+    }
+}
+
+/// Round parameters
+#[derive(Debug, Clone)]
+pub struct Round {
+    /// Proof counter
```
djetchev, line 184: Idem: for this structure (see Setup)

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Round candidate hash mapped to [1, n_p]
+    h_usize: usize,
+    /// Approximate size of set Sp to lower bound
+    n_p: usize,
+}
+
+impl Round {
+    /// Oracle producing a uniformly random value in [1, n_p] used for round candidates
+    /// We also return hash(data) to follow the optimization presented in Section 3.3
+    fn h1(data: Vec<Vec<u8>>, n_p: usize) -> (Hash, usize) {
```
djetchev, line 201: function parameter - why not borrowing the data here via reference to immutable slice 
```
fn h1(data: &[Vec<u8>], ...)
```

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+
+/// Setup input parameters
+#[derive(Debug, Clone)]
+pub struct Params {
+    /// Soundness security parameter
+    pub lambda_sec: usize,
+    /// Completeness security parameter
+    pub lambda_rel: usize,
+    /// Approximate size of set Sp to lower bound
+    pub n_p: usize,
```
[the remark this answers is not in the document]
rrtoledo: Lambda technically is a number of bits so it is an integer.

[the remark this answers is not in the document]
rrtoledo: As for np, because it is used when indexing the bins, we get errors when changing to u64
```
error[E0277]: the type `[Vec<[u8; 32]>]` cannot be indexed by `u64`
   --> src/centralized.rs:299:24
    |
299 |         for &s in bins[round.h_usize].iter() {
    |                        ^^^^^^^^^^^^^ slice indices are of type `usize` or ranges of `usize`
    |
```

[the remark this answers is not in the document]
rrtoledo: Pushed changes for lambda and r

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+    fn new() -> Self {
+        Proof {
+            r: 0,
+            d: 0,
+            items: vec![],
+        }
+    }
+
+    /// Oracle producing a uniformly random value in [1, n_p] used for prehashing S_p
+    fn h0(setup: &Setup, v: usize, s: Data) -> usize {
```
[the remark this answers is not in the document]
rrtoledo: Just added some. I initially wanted to have a 2 values: one hardwritten in the code and the second provided by the user. What do you think?

src/centralized.rs
```diff
@@ -0,0 +1,504 @@
...
+            limit = l;
+        }
+        (limit, None)
+    }
+
+    /// Alba's proving algorithm, based on a depth-first search algorithm.
+    /// Calls up to setup.r times the prove_index function and returns an empty
+    /// proof if no suitable candidate is found.
+    pub fn prove(setup: &Setup, set: &[Element]) -> Self {
+        for v in 0..setup.r {
```
tolikzinovyev, line 346: As mentioned in https://github.com/cardano-scaling/alba/pull/34#discussion_r1813060453, we should probably cut down `set` to `2 * n_p` elements to make sure the prover running time stays low.

src/centralized.rs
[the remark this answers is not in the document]
tolikzinovyev: I don't agree about lambda. The number of what bits? Suppose the prover needs to succeed only with probability 1/4. Then it is more efficient to set `lambda_rel` to log(4/3) instead of 1.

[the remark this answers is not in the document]
tolikzinovyev: If we make `n_p` `u64`, even though it will not let 32-bit provers generate a proof, 32-bit verifiers could still verify a certificate. It could in principle be useful. In the prover code we can just convert to `n_p` to `usize` for indexing into a slice. We don't even need to check that `n_p` fits in `usize` since if `set` really contains `n_p` elements, it must!

Cargo.toml
```diff
@@ -1,9 +1,29 @@
...
 description = "A Rust implementation of Approximate Lower Bound Arguments (ALBAs)."
 categories = ["cryptography"]
 include = ["**/*.rs", "Cargo.toml", "README.md", ".gitignore"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+blake2 = "0.10.6"
+rand_core = "0.6.4"
+rayon = "1.10.0"
```
tolikzinovyev, line 14: please remove this as well

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+
+const DATA_LENGTH: usize = 32;
+const DIGEST_SIZE: usize = 32;
+
+type Data = [u8; DATA_LENGTH];
+type Hash = [u8; DIGEST_SIZE];
+
+/// Setup input parameters
+#[derive(Debug, Clone)]
+pub struct Params {
```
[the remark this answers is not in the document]
rrtoledo: I was thinking of having two separate structures for decentralise and centralise as decentralise also needs mu. As such, centralise parameters are only used in this one file.

src/centralized.rs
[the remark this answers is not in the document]
rrtoledo: In that case, this is for another issue. I'll mark this comment  as resolve if that's fine with you.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Oracle producing a uniformly random value in [1, n_p] used for prehashing S_p
+    fn h0(setup: &Setup, v: usize, s: Data) -> usize {
+        let mut data = vec![v.to_ne_bytes().to_vec()];
+        data.push(s.to_vec());
+        let digest = utils::combine_hashes::<DIGEST_SIZE>(data);
+        utils::oracle_uniform(&digest, setup.n_p)
+    }
+
+    /// Oracle defined as Bernoulli(q) returning 1 with probability q and 0 otherwise
+    fn h2(setup: &Setup, r: &Round) -> bool {
```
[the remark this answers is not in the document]
rrtoledo: Good idea.
This was not in the paper btw!

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            if let (_, Some(proof)) = Proof::prove_index(setup, set, v + 1) {
+                return proof;
+            }
+        }
+        Proof::new()
+    }
+
+    /// Alba's proving algorithm used for benchmarking, returning a proof as
+    /// well as the number of  steps ran to find it.
+    pub fn bench(setup: &Setup, set: &[Data]) -> (usize, usize, Self) {
```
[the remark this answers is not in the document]
rrtoledo: Can you explain me more how to have the internal prove benched if it is not public?

As for calling the prove function on the benched, I have no problem with it, I just wanted to avoid the extra accumulator overhead.

src/centralized.rs
rrtoledo: Removed it.
Pyrrhos had sent a summary of the paper for the parameter choice and it was to check whether this corresponded with the table borns ( < lamba, > lambda^2, ...)

src/utils.rs
rrtoledo: The "inverse proba" was because of an error of mine when I used this oracle for H2 with parameter 1/q. Updated.

rrtoledo: What is wrong with the function name?

src/centralized.rs
rrtoledo: Renamed v,t, for the time being.
Opened new issue for more descriptive names

rrtoledo: Moved to new issue https://github.com/cardano-scaling/alba/issues/38

src/centralized.rs
rrtoledo: Sure.
Nitpicking: we could argue that the lottery scheme is both centralized and decentralized hence there is only one centralized scheme.

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            }
+            w
+        }
+
+        let n_p_f64 = params.n_p as f64;
+        let n_f_f64 = params.n_f as f64;
+        let lognpnf = (n_p_f64 / n_f_f64).log2();
+        let lambda_rel = params.lambda_rel as f64;
+        let logrel = lambda_rel.log2();
+        let lambda_sec = (params.lambda_sec as f64) + logrel;
```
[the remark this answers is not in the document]
rrtoledo: That's from the paper section 3.2.2
```
 we set r := λrel,
and to compensate for the loss in soundness, we set λsec := λsec + log λrel. 
```
It needs to be updated to be added only in the small case though!

src/centralized.rs
```diff
@@ -0,0 +1,499 @@
...
+            let d = (16.0 * u_f64 * lbar).ceil();
+            assert!(n_p_f64 >= d * d / (9.0 * lbar));
+
+            let w = compute_w(u_f64, lambda_rel1);
+            Setup {
+                n_p: params.n_p,
+                u,
+                r: (lambda_rel / lambda_rel1).ceil() as usize,
+                d: d as usize,
+                q: (2.0 * lbar / d).recip(),
```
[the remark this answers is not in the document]
rrtoledo: Well spotted! That's an error from using the same oracle that was not fully fixed.

src/centralized.rs
rrtoledo: See issue https://github.com/cardano-scaling/alba/issues/38

src/centralized.rs
rrtoledo: @djetchev As I removed the parallelization (to be treated in its own issue) I have updated the code similarly to your idea but with a fold.

src/centralized.rs
rrtoledo: I hear you.

The problem I have with this is that I would have to initialize the array with setup.n_p and as it is not constant, it is not (easily perhaps) possible. 

Another problem is that Vector does not implement the Copy trait, as such to initialise bins I would have to do something like: `let bins : [Vec<Element>; N] = vec![vec![]; N].try_into().unwrap();`

src/centralized.rs
rrtoledo: See issue https://github.com/cardano-scaling/alba/issues/38

src/utils.rs
rrtoledo: I do not need to.
That's what I was doing.
I updated the code with a iter().foreach() in the end

src/utils.rs
rrtoledo: I left the code as is (I think that was from Arnaud) as I was not sure what to choose here as parameter. I think the best would be to take as input lambda and use this.

src/utils.rs
rrtoledo: Fill free to open an issue to reorder the code after it's merged

src/utils.rs
rrtoledo: I kept the code as is from Arnaud. I think it was named so to remind that the bytes are somewhat random

Cargo.toml
```diff
@@ -1,9 +1,28 @@
...
+
+[[bench]]
+name = "centralized_telescope_time"
+harness = false
+path = "benches/centralized_telescope.rs"
+
+[[bench]]
+name = "centralized_telescope_step"
+harness = false
+path = "benches/centralized_telescope.rs"
```
djetchev, line 28: newline - we can stick to having the end of file being a newline always

src/lib.rs
tolikzinovyev: Do we actually need a description of alba here? We have README, the website and @curiecrypt is working on more documentation.

src/centralized.rs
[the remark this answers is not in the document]
rrtoledo: In that case why not lambda as f32? Because of the precision? We should not need or want too large values for it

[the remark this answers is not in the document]
tolikzinovyev: Yes, it doesn't hurt to have more precision.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// Round candidate hash
+    h: Hash,
+    /// Round candidate hash mapped to [1, n_p]
+    h_u64: u64,
+    /// Approximate size of set Sp to lower bound
+    n_p: u64,
+}
+
+impl Round {
+    /// Oracle producing a uniformly random value in [1, n_p] used for round candidates
```
djetchev, line 194: @rrtoledo - there might be a way to improve the memory usage in this function as follows: instead of creating the intermediate `data` (a structure that is `Vec<Vec<>>`) and use memory for it, you can use the `hasher` directly and call `hasher.update()` in-place. That will save you memory - something like this: 
```
fn h1(input: &[Vec<u8>], n_p: u64) -> (Hash, u64) {
    let mut hasher = Blake2b::new();
    hasher.update(b"Telescope-H1");
    for i in input {
        hasher.update(i);
    }
    let digest: Hash = hasher.finalize().try_into().unwrap();
    (digest, utils::sample_uniform(&digest, n_p))
}
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// We also return hash(data) to follow the optimization presented in Section 3.3
+    fn h1(input: &[Vec<u8>], n_p: u64) -> (Hash, u64) {
```
djetchev, line 196: @rrtoledo - there might be a way to improve the memory usage in this function as follows: instead of creating the intermediate `data` (a structure that is `Vec<Vec<>>`) and use memory for it, you can use the `hasher` directly and call `hasher.update()` in-place. That will save you memory - something like this: 
```
fn h1(input: &[Vec<u8>], n_p: u64) -> (Hash, u64) {
    let mut hasher = Blake2b::new();
    hasher.update(b"Telescope-H1");
    for i in input {
        hasher.update(i);
    }
    let digest: Hash = hasher.finalize().try_into().unwrap();
    (digest, utils::sample_uniform(&digest, n_p))
}
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+            n_p,
+        }
+    }
+
+    /// Updates a round with an element of S_p
+    /// Replaces the hash $h$ with $h' = H1(h, s)$ and the random value as oracle(h', n_p)
+    pub fn update(r: &Round, s: Element) -> Round {
+        let mut s_list = r.s_list.clone();
+        s_list.push(s);
+        let mut data = vec![r.h.clone().to_vec()];
```
djetchev, line 226: as before, if you want to avoid creating this new data and cloning stuff, you can use the `Blake2b::update()`. That way, you can directly use the data available in `r` to update the hasher (same idea as the comment in line 196.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+        let mut data = vec!["Telescope-H0".as_bytes().to_vec()];
+        data.push(v.to_ne_bytes().to_vec());
+        data.push(s.to_vec());
+        let digest = utils::combine_hashes::<DIGEST_SIZE>(&data);
+        utils::sample_uniform(&digest, setup.n_p)
+    }
+
+    /// Oracle defined as Bernoulli(q) returning 1 with probability q and 0 otherwise
+    fn h2(setup: &Setup, r: &Round) -> bool {
+        let mut data = vec!["Telescope-H2".as_bytes().to_vec()];
```
djetchev, line 263: idem - you can do this without the intermediate allocations and copies.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// and returns first round candidate Round{t, x_1, ..., x_u)} such that:
+    /// - for all i ∈ [0, u-1], H0(x_i+1) ∈ bins[H1(t, x_1, ..., x_i)]
+    /// - H2(t, x_0, ..., x_u) = true
+    fn dfs(
+        setup: &Setup,
+        bins: &[Vec<Element>],
+        round: &Round,
+        limit: u64,
+    ) -> (u64, Option<Proof>) {
+        if round.s_list.len() as u64 == setup.u {
```
djetchev, line 279: lines 279--288: apparently Rust allows you to use a return statement with ternary operator
```
return if Proof::h2(setup, round) {
    (limit, Some(Proof {
         v: round.v,
          t: round.t,
          items: round.s_list.clone(),
     }))
} else {
    (limit, None)
};
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+                let items = round.s_list.clone();
+                return (limit, Some(Proof { v, t, items }));
+            } else {
+                return (limit, None);
+            }
+        }
+
+        bins[round.h_u64 as usize]
+            .iter()
+            .fold((limit, None), |(l, proof_opt), &s| {
```
djetchev, line 292: An alternative to the use of `fold()` with the described closure is to simply use a `for` loop over `s`.  This will make the code more readable but also has several efficiency advantages: 

1. With the `fold()` function and the closure, you are not allowing for potential early returns when a proof is found. 

2. You are doing recursive calls on every iteration, thus, leading to deep call stacks. 

3. You keep an accumulator `(l, proof_opt)` on each step which is an overhead compared to a mutation in the `for` loop

4. the calls to `Round::update()` might lead to unnecessary clones.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+            } else {
+                let (l, opt) = Proof::prove_index(setup, set, v + 1);
+                (limit + l, r, opt)
+            }
+        })
+    }
+
+    /// Alba's verification algorithm, follows proving algorithm by running the
+    /// same depth-first search algorithm.
+    pub fn verify(setup: &Setup, proof: Proof) -> bool {
```
djetchev, line 341: Is there a reason to not take proof as a reference, i.e., `proof: &Proof`?

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+        if proof.t >= setup.d || proof.v >= setup.r || proof.items.len() as u64 != setup.u {
+            return false;
+        }
+        let r0 = Round::new(proof.v, proof.t, setup.n_p);
+        let (b, round) = proof.items.iter().fold((true, r0), |(b, r), &s| {
```
djetchev, line 346: idem - instead of `fold`, you can use a `for` loop to both improve readability of the code but also allow for early termination (instead of keeping a boolean in the accumulator). Simply return `false` at the iteration when you find a `h0` mismatch and return early.

src/centralized_telescope.rs
tolikzinovyev: @djetchev I'm planning to solve this using iterators when working on https://github.com/cardano-scaling/alba/issues/36, so I think we can leave it as is for now.

I think a bigger performance issue is copying the elements list for every DFS step -- it's an asymptotic slowdown, not just a constant factor one. But I'm fine to leave it for a future PR.

src/centralized_telescope.rs
rrtoledo: I saw. I first wrote somethine like the following but it wasn't wasy to read 
```return (limit, if Proof::h2(setup, round) {
    (limit, Some(Proof {
         v: round.v,
          t: round.t,
          items: round.s_list.clone(),
     })) else { None })
};```
I'll update with yours!

Comment (tolikzinovyev) [reply src/centralized_telescope.rs:279]: Another option:
```
let proof = if Proof::h2(setup, round) {
    Some(Proof {
         v: round.v,
          t: round.t,
          items: round.s_list.clone(),
     })
} else {
    None
};
return (limit, proof)
```

File: benches/centralised_repetition.rs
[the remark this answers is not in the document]
Comment (rrtoledo) [reply benches/centralised_repetition.rs:1]: Covered in https://github.com/cardano-scaling/alba/issues/7

File: src/centralized.rs (hunk 42 above)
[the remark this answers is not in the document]
Comment (rrtoledo) [reply src/centralized.rs]: Creating new issue for this https://github.com/cardano-scaling/alba/issues/46

File: src/utils.rs (hunk 21 above)
Comment (rrtoledo) [reply src/utils.rs]: Covered in  [     - ](https://github.com/cardano-scaling/alba/issues/46)

File: src/centralized_telescope.rs (hunk 49 above)
Comment (rrtoledo) [reply src/centralized_telescope.rs:196]: Updated to using blake2s directly before #36

File: src/lib.rs (hunk 26 above)
Comment (rrtoledo) [reply src/lib.rs:4]: Let's agree with @tolikzinovyev and keep all documentation in the README and website.

File: src/utils.rs (hunk 16 above)
Comment (rrtoledo) [reply src/utils.rs:80]: Sure

**Comment by rrtoledo** [comment]
@curiecrypt @djetchev @tolikzinovyev I have now answered all conversations (the ones marked as unresolved where answered but I forgot to click the "resolve" button and I made a force push a long time ago so they are not accessible anymore sorry).

File: src/centralized_telescope.rs [hunk 56]
```diff
@@ -0,0 +1,419 @@
...
+            }
+        }
+
+        let mut limit = 0;
+        for t in 0..setup.d {
+            if limit == setup.b {
+                return (limit, None);
+            }
+            if let Some(r) = Round::new(v, t, setup.n_p, setup.sec_param) {
+                let (l, proof_opt) = Self::dfs(setup, &bins, &r, limit.saturating_add(1));
```
curiecrypt, line 301: Out of curiosity, what odds do we have an overflow with this addition?
Same for the other occurrences of `saturating`

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// Indexed proving algorithm, returns an empty proof if no suitable
+    /// candidate is found within the setup.b steps.
+    fn prove_index(setup: &Setup, set: &[Element], v: u64) -> (u64, Option<Self>) {
+        let mut bins: Vec<Vec<Element>> = vec![vec![]; setup.n_p as usize];
+        for &s in set {
+            match Self::h0(setup, v, s) {
+                Some(h) => {
+                    bins[h as usize].push(s);
+                }
+                None => return (0, None),
```
curiecrypt, line 291: Future reminder: When we implement the `errors`, we definitely should specify in which case we return `none`.
Also, it could be useful to explain the difference between returning an empty proof or `none` in the docs.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+        hasher.update(r.h);
+        let digest: Hash = hasher.finalize().into();
+        utils::sample_bernouilli(&digest, setup.q, setup.sec_param)
+    }
+
+    /// Depth-first search which goes through all potential round candidates
+    /// and returns first round candidate Round{t, x_1, ..., x_u)} such that:
+    /// - for all i ∈ [0, u-1], H0(x_i+1) ∈ bins[H1(t, x_1, ..., x_i)]
+    /// - H2(t, x_0, ..., x_u) = true
+    fn dfs(setup: &Setup, bins: &[Vec<Element>], round: &Round, limit: u64) -> (u64, Option<Self>) {
```
curiecrypt, line 255: !Future!
I remember that I commented that we should comply with the names in the paper.
However, `limit` does not seem to cover its actual functionality.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+
+const DATA_LENGTH: usize = 32;
+const DIGEST_SIZE: usize = 32;
+
+type Element = [u8; DATA_LENGTH];
+type Hash = [u8; DIGEST_SIZE];
+
+/// Setup input parameters
+#[derive(Debug)]
+pub struct Params {
```
curiecrypt, line 16: What if we define `lambda_sec`, `lambda_rel`, and `sec_param` as global variables?

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// Proof max counter
+    pub r: u64,
+    /// Proof max 2nd counter
+    pub d: u64,
+    /// Probability q
+    pub q: f64,
+    /// Computation bound
+    pub b: u64,
+}
+impl Setup {
```
curiecrypt, line 45: This version of Setup seems much better and more user-friendly.
However, I would prefer to see the `new` function at the bottom.

curiecrypt (dismissed):
Just minor comments.  Mostly, could be addressed after we merge this PR.
One important point is to revisit the visibility of structs, variables, and functions. This could also be considered after this first version.
Thanks for your effort @rrtoledo 🙏 
LGTM 👍

src/centralized_telescope.rs
rrtoledo: I think this relates to the question of the code purpose.
If this  code is meant to be used as reference, we should leave the lambdas as parameters.
However if it is supposed to be a product, our "customers" would be developpers and we should only provide functions with set security. 
What is the code (reference/product) has not been settled, so right now I think it is reasonable to leave the lambdas as variables. We could however also provide functions with set security parameters in case.

src/centralized_telescope.rs
rrtoledo: Yes, you wanted to change the name from "steps" to "limit". I had argued that here we are incrementing the value and not decreasing it contrary to the paper. Perhaps we can add this in the variable name issue?

src/centralized_telescope.rs
rrtoledo: Sure we can be more verbose when using errors (here it is the initialisation that failed for instance, not the proof generation per se).
We do not have empty proofs anymore by the way!

src/centralized_telescope.rs
rrtoledo: Considering that the maximum bound is O(lambda^3) and that the security parameters are usually set at 128-256bits of security, we can expect the maximum bounds to vary around 2**{21-24}.
The variable `limit` is a u64., that is 2**64. The odds should be quite low, even if the security parameters were to drasticaly increase. :)

src/centralized_telescope.rs
rrtoledo: I am not sure to understand what you mean? The `new` function is at the bottom (of the impl Setup), do you mean you want the implementation of Setup to be at the bottom (first we define all structures then the implementations)?
This should be covered in https://github.com/cardano-scaling/alba/issues/46, where we will split the file depending on the structure.

src/centralized_telescope.rs
djetchev: I have a slight preference for having the security parameters as variables too and maybe suggest some default values though those values will be different for the different applications.

src/centralized_telescope.rs
curiecrypt: Ooops! Yes, I missed it. But, agreed on being more verbose

src/centralized_telescope.rs
djetchev: I think Gamze means that the `new()` function should be defined first in the `impl Setup` block

curiecrypt: nope. Not at the bottom, but at the top, I would say. And, should be addressed in #46 for sure

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+        hasher.update(second_input);
+        let digest: Hash = hasher.finalize().into();
+        (digest, utils::sample_uniform(&digest, n_p, sec_param))
+    }
+
+    /// Output a round from a proof counter and n_p
+    /// Initilialises the hash with H1(t) and random value as oracle(H1(t), n_p)
+    pub fn new(v: u64, t: u64, n_p: u64, sec_param: u64) -> Option<Self> {
+        let (h, h_u64_opt) = Self::h1(
+            v.to_be_bytes().as_ref(),
```
djetchev, line 188: Why not use stack-allocated byte arrays 
```
let mut v_bytes = [0u8; 8];
let mut t_bytes = [0u8; 8];
```

rrtoledo: Do you mean something like this?
```
let v_bytes : [u8; 8] = v.to_be_bytes();
        let t_bytes : [u8; 8] = t.to_be_bytes();
        let (h, h_u64_opt) = Self::h1(
            v_bytes.as_ref(),
            t_bytes.as_ref(),
            n_p,
            sec_param,
        );
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,419 @@
...
+            t.to_be_bytes().as_ref(),
+            n_p,
+            sec_param,
+        );
+        h_u64_opt.map(|h_u64| Self {
+            v,
+            t,
+            s_list: vec![],
```
djetchev, line 197: There is a subtle difference between using `vec![]` and `Vec::new()` - the latter uses no macro expansion while the former expands to something like this :) 
```
macro_rules! vec {
    () => {
        {
            let mut v = Vec::new();
            v
        }
    };
    ($elem:expr; $n:expr) => { /* ... */ };
    ($($x:expr),+ $(,)?) => { /* ... */ };
}
```
Thus, using `Vec::new()` here is slightly more efficient in terms of code to process by the compiler, but the runtime difference is negligible, so primarily a matter of style :)

src/centralized_telescope.rs
rrtoledo: Updated

src/centralized_telescope.rs
rrtoledo: Updated

tolikzinovyev:
> @curiecrypt @djetchev @tolikzinovyev I have now answered all conversations (the ones marked as unresolved where answered but I forgot to click the "resolve" button and I made a force push a long time ago so they are not accessible anymore sorry).

The unresolved threads can be accessed here.
Screenshot 2024-10-30 at 13-42-02 Adding centralised alba (bounded scheme) by rrtoledo · Pull Request #33 · cardano-scaling_alba

tolikzinovyev (commented):
I think we are getting closer to merging this. The parameter derivation code definitely improved! I'm fine with postponing code reorganization as long as it is the next immediate PR. I will do careful review of parameter derivation and the verifier code after refactoring, and prover code after optimizing DFS, so that I don't have to do it twice.

The remaining things to fix is the limit logic in DFS and various small issues. I also need to review the test code, but I don't expect big issues there.

rrtoledo:
> > @curiecrypt @djetchev @tolikzinovyev I have now answered all conversations (the ones marked as unresolved where answered but I forgot to click the "resolve" button and I made a force push a long time ago so they are not accessible anymore sorry).
> 
> The unresolved threads can be accessed here. Screenshot 2024-10-30 at 13-42-02 Adding centralised alba (bounded scheme) by rrtoledo · Pull Request #33 · cardano-scaling_alba

If I click on one of the old discussion, it redirects me to an error page
```
We went looking everywhere, but couldn’t find those commits.
Sometimes commits can disappear after a force-push. Head back to the [latest changes here](https://github.com/cardano-scaling/alba/pull/33/files).
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            r: (params.lambda_rel / lambda_rel1).ceil() as u64,
+            d: d as u64,
+            q: 2.0 * lbar_over_d,
+            b: ((w * lbar_over_d + 1.0) * exponential * d * u_f64 + d).floor() as u64,
+        }
+    }
+
+    /// Setup algorithm taking a Params as input and returning setup parameters (u,d,q)
+    pub fn new(params: &Params) -> Self {
+        let sec_param = params.lambda_rel.max(params.lambda_sec).ceil() as u64;
```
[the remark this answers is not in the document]
rrtoledo: Another problem, if we use lambda directly it may overflow:
```
        let epsilon_fail: u64 = 1 << sec_param;
```
Perhaps we should indeed just hardcode a value like before.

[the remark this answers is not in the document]
rrtoledo: c.f. previous discussion

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+
+#[derive(Debug, Clone)]
+/// Alba proof
+pub struct Proof {
+    /// Proof counter
+    v: u64,
+    /// Proof 2nd counter
+    t: u64,
+    /// Proof tuple
+    items: Vec<Element>,
```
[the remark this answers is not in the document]
rrtoledo: That will be updated in https://github.com/cardano-scaling/alba/issues/38

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            } else {
+                None
+            };
+            return (limit, proof_opt);
+        }
+
+        let mut l = limit;
+        for &s in &bins[round.h_u64 as usize] {
+            if let Some(r) = Round::update(round, s, setup.sec_param) {
+                let (l_dfs, proof_opt) = Self::dfs(setup, bins, &r, l.saturating_add(1));
```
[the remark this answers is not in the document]
rrtoledo: C.f. @curiecrypt previous review, this was first called the number of steps (as it is interesting for benchmark purposes) and renamed limit to relate to the paper.
Feel free to suggest a better name in https://github.com/cardano-scaling/alba/issues/38

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+        hasher.update(r.h);
+        let digest: Hash = hasher.finalize().into();
+        utils::sample_bernouilli(&digest, setup.q, setup.sec_param)
+    }
+
+    /// Depth-first search which goes through all potential round candidates
+    /// and returns first round candidate Round{t, x_1, ..., x_u)} such that:
+    /// - for all i ∈ [0, u-1], H0(x_i+1) ∈ bins[H1(t, x_1, ..., x_i)]
+    /// - H2(t, x_0, ..., x_u) = true
+    fn dfs(setup: &Setup, bins: &[Vec<Element>], round: &Round, limit: u64) -> (u64, Option<Self>) {
```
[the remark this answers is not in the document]
rrtoledo: I must have forgot it when changing the fold to a for loop! Well spoted

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+        (l, None)
+    }
+
+    /// Indexed proving algorithm, returns an empty proof if no suitable
+    /// candidate is found within the setup.b steps.
+    fn prove_index(setup: &Setup, set: &[Element], v: u64) -> (u64, Option<Self>) {
+        let mut bins: Vec<Vec<Element>> = Vec::with_capacity(setup.n_p as usize);
+        for _ in 0..setup.n_p {
+            bins.push(Vec::new());
+        }
```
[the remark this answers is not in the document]
rrtoledo: Dimitar wanted not to use the macro

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+        for &s in set {
+            match Self::h0(setup, v, s) {
+                Some(h) => {
+                    bins[h as usize].push(s);
+                }
+                None => return (0, None),
+            }
+        }
+
+        let mut limit = 0;
```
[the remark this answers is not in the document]
rrtoledo: You mean `setup.b`? C.f. previous comment

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+        (limit, None)
+    }
+
+    /// Alba's proving algorithm, based on a depth-first search algorithm.
+    /// Calls up to setup.r times the prove_index function and returns an empty
+    /// proof if no suitable candidate is found.
+    pub fn prove(setup: &Setup, set: &[Element]) -> Option<Self> {
+        // Take only up to 2*np elements for efficiency
+        let two_np = setup.n_p.saturating_mul(2) as usize;
+        let truncated_set = if set.len() >= two_np {
```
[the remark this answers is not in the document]
rrtoledo: What about truncating in prove index directly by doing:
```
        for &s in set.iter().take((2u64 * setup.n_p) as usize) {
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            &set.iter().take(two_np).copied().collect::<Vec<Element>>()
```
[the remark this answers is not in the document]
rrtoledo: c.f. discussion above

src/utils.rs
```diff
@@ -0,0 +1,78 @@
...
+    // Output i in [0; y-1] from hash
+    debug_assert!(8.0 * hash.len() as f32 >= (y as f32).log2());
+    let i = from_bytes_le(hash) & y.saturating_sub(1);
+    // Return true if i < x
+    i < x
+}
+
+// Returns the integer representation of, up to the 8 first bytes of, the
+// input bytes in little endian
+fn from_bytes_le(bytes: &[u8]) -> u64 {
```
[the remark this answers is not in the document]
rrtoledo: These methods take as input `[u8;8]` hence the wrapper

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Round candidate hash mapped to [1, n_p]
+    h_u64: u64,
+    /// Approximate size of set Sp to lower bound
+    n_p: u64,
+}
+
+impl Round {
+    /// Oracle producing a uniformly random value in [1, n_p] used for round candidates
+    /// We also return hash(data) to follow the optimization presented in Section 3.3
+    fn h1(
```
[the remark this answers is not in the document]
rrtoledo: Let's add this to https://github.com/cardano-scaling/alba/issues/36 as the domain separation  may be one more variable in the Hx structure

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            s_list: Vec::new(),
+            h,
+            h_u64,
+            n_p,
+        })
+    }
+
+    /// Updates a round with an element of S_p
+    /// Replaces the hash $h$ with $h' = H1(h, s)$ and the random value as oracle(h', n_p)
+    pub fn update(r: &Self, s: Element, sec_param: u64) -> Option<Self> {
```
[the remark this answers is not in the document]
rrtoledo: Good idea, let's add it to https://github.com/cardano-scaling/alba/issues/36

src/utils.rs
```diff
@@ -0,0 +1,78 @@
...
+    // power of two. *(up to 8 bytes, in little endian)
+    fn mod_non_power_of_2(hash: &[u8], n: u64, sec_param: u64) -> Option<u64> {
+        fn log_base2(x: u64) -> u64 {
+            u64::from(
+                u64::BITS
+                    .saturating_sub(x.leading_zeros())
+                    .saturating_sub(1),
+            )
+        }
+        let epsilon_fail: u64 = 1 << sec_param;
```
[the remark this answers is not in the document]
rrtoledo: Issue created

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Oracle producing a uniformly random value in [1, n_p] used for round candidates
+    /// We also return hash(data) to follow the optimization presented in Section 3.3
+    fn h1(
+        first_input: &[u8],
+        second_input: &[u8],
+        n_p: u64,
+        sec_param: u64,
+    ) -> (Hash, Option<u64>) {
+        let mut hasher = Blake2s256::new();
+        hasher.update(b"Telescope-H1");
```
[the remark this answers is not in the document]
rrtoledo: Sure, let's add it to https://github.com/cardano-scaling/alba/issues/36 to set some requirements on the user's provided hash domain separation

src/utils.rs
rrtoledo: Let's change the name in https://github.com/cardano-scaling/alba/issues/38

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+        (0..setup.r).find_map(|v| Self::prove_index(setup, set, v).1)
+    }
+
+    /// Alba's verification algorithm, follows proving algorithm by running the
+    /// same depth-first search algorithm.
+    pub fn verify(setup: &Setup, proof: &Self) -> bool {
+        if proof.t >= setup.d || proof.v >= setup.r || proof.items.len() as u64 != setup.u {
+            return false;
+        }
+        if let Some(mut round) = Round::new(proof.v, proof.t, setup.n_p, setup.sec_param) {
```
tolikzinovyev, line 336: This function has a lot of indentation which makes reading it hard. Let's instead short-circuit just like the above three lines.
```suggestion
        let mut round = match Round::new(proof.v, proof.t, setup.n_p, setup.sec_param) {
          Some(x) => x,
          None => return false,
        };
```
The same can be done in a couple more places below.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    use rand_core::{RngCore, SeedableRng};
+
+    #[test]
+    fn test_verify() {
+        let mut rng = ChaCha20Rng::from_seed(Default::default());
+        let nb_tests = 1_000;
+        let set_size = 1_000;
+        for _t in 0..nb_tests {
+            let seed = rng.next_u32().to_be_bytes().to_vec();
+            let s_p = gen_items::<DATA_LENGTH>(&seed, set_size);
```
tolikzinovyev, line 372: why not just pass `rng` to `gen_items` and let it use it instead of hashing?

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            };
+            let setup = Setup::new(&params);
+            let proof = Proof::prove(&setup, &s_p).unwrap();
+            assert!(Proof::verify(&setup, &proof.clone()));
+            let proof_d = Proof {
+                v: proof.v,
+                t: proof.t.wrapping_add(1),
+                items: proof.items.clone(),
+            };
+            assert!(!Proof::verify(&setup, &proof_d));
```
tolikzinovyev, line 387: Suggest putting each bad proof in its own scope. This also solves the problem with having to come up with new variable names.
```suggestion
            {
              let proof = Proof {
                  v: proof.v,
                  t: proof.t.wrapping_add(1),
                  items: proof.items.clone(),
              };
              assert!(!Proof::verify(&setup, &proof));
            }
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::gen_items;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::{RngCore, SeedableRng};
+
+    #[test]
+    fn test_verify() {
+        let mut rng = ChaCha20Rng::from_seed(Default::default());
```
tolikzinovyev, line 367: This is deterministic, right?

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+            wrong_items.push(last_item);
+            let proof_itembis = Proof {
+                v: proof.v,
+                t: proof.t,
+                items: wrong_items.clone(),
+            };
+            assert!(!Proof::verify(&setup, &proof_itembis));
+        }
+    }
+}
```
tolikzinovyev, line 421: Do we have an issue for (basic) tests? I think we could benefit from testing the DFS bound logic, parameter derivation, and maybe some other tests.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    #[test]
+    fn test_verify() {
+        let mut rng = ChaCha20Rng::from_seed(Default::default());
+        let nb_tests = 1_000;
+        let set_size = 1_000;
+        for _t in 0..nb_tests {
+            let seed = rng.next_u32().to_be_bytes().to_vec();
+            let s_p = gen_items::<DATA_LENGTH>(&seed, set_size);
+            let params = Params {
```
tolikzinovyev, line 373: maybe move it up to where the other constants are?

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    use super::*;
+    use crate::test_utils::gen_items;
+    use rand_chacha::ChaCha20Rng;
+    use rand_core::{RngCore, SeedableRng};
+
+    #[test]
+    fn test_verify() {
+        let mut rng = ChaCha20Rng::from_seed(Default::default());
+        let nb_tests = 1_000;
+        let set_size = 1_000;
```
tolikzinovyev, line 369: we don't need this many since `prove()` truncates to 160 anyway

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Round candidate hash
+    h: Hash,
+    /// Round candidate hash mapped to [1, n_p]
+    h_u64: u64,
+    /// Approximate size of set Sp to lower bound
+    n_p: u64,
+}
+
+impl Round {
+    /// Oracle producing a uniformly random value in [1, n_p] used for round candidates
```
tolikzinovyev, line 167:
```suggestion
    /// Oracle producing a uniformly random value in [0, n_p) used for round candidates
```

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Proof counter
+    v: u64,
+    /// Proof 2nd counter
+    t: u64,
+    /// Proof tuple
+    items: Vec<Element>,
+}
+
+impl Proof {
+    /// Oracle producing a uniformly random value in [1, n_p] used for prehashing S_p
```
tolikzinovyev, line 228:
```suggestion
    /// Oracle producing a uniformly random value in [0, n_p) used for prehashing S_p
```

tolikzinovyev (commented):
I think `limit` and `sec_param` should be renamed now since they are very confusing and it's no longer just a matter of style / preference, but I'm ok to leave it for https://github.com/cardano-scaling/alba/issues/38 as long as it is addressed soon.

src/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: I saw you commented on https://github.com/cardano-scaling/alba/issues/38, so let's discuss there.

src/centralized_telescope.rs
rrtoledo: I believe it is

src/centralized_telescope.rs
rrtoledo: I forgot to change this, we often used np and nf as percentages before!

src/centralized_telescope.rs
rrtoledo: I prefer using separating these functions, it's cleaner: we iterate the rng per test iteration.

src/centralized_telescope.rs
rrtoledo: Sure

src/centralized_telescope.rs
rrtoledo: Sure, feel free to add more tests in a separate issue: https://github.com/cardano-scaling/alba/issues/49

src/centralized_telescope.rs
rrtoledo: That does not bring much. Feel free to do this in the unit test issue I assigned to you: https://github.com/cardano-scaling/alba/issues/49

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    /// Alba's proving algorithm, based on a depth-first search algorithm.
+    /// Calls up to setup.r times the prove_index function and returns an empty
+    /// proof if no suitable candidate is found.
+    pub fn prove(setup: &Setup, set: &[Element]) -> Option<Self> {
+        (0..setup.r).find_map(|v| Self::prove_index(setup, set, v).1)
+    }
+
+    /// Alba's verification algorithm, follows proving algorithm by running the
```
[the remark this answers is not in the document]
rrtoledo: Updated

src/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: I find it cleaner as we only use the variable `set` in one place.

src/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: Both lines are readable. The current solution is even a bit faster. Let's keep it.

src/centralized_telescope.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Oracle defined as Bernoulli(q) returning 1 with probability q and 0 otherwise
+    fn h2(setup: &Setup, r: &Round) -> bool {
+        let mut hasher = Blake2s256::new();
+        hasher.update(b"Telescope-H2");
+        hasher.update(r.h);
+        let digest: Hash = hasher.finalize().into();
+        utils::sample_bernouilli(&digest, setup.q, setup.sec_param)
+    }
+
+    /// Depth-first search which goes through all potential round candidates
```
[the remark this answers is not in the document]
rrtoledo: Not sure why it doesn't appear here but it was updated (check the comment in the last commit you'll see).

src/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: Removed.

djetchev (approved):
Great job, @rrtoledo ! I know it was a lot of work and I am sure we will further improve it but already quite happy with it. 

LGTM!

PR #34, opened 2024-10-16
Cargo documentation for centralized version

curiecrypt:
## Content

This is the first draft of the comprehensive documentation for the **Telescope-ALBA protocol**.  The scope and content will differ from the [existing documentation](https://alba.cardano-scaling.org/) currently available.  

The goal of this document is to bridge **theory and practical implementation** by providing a seamless connection between **code documentation** (`rust doc`) and **user-friendly explanations** based on the paper.  

This draft was generated using the [cargo doc](https://doc.rust-lang.org/cargo/commands/cargo-doc.html) tool. Once published online, the output could serve as an intuitive interface for the library. (See [this](https://doc.dalek.rs/bulletproofs/index.html).)

This PR includes:
- A crate called `cargodocs` to operate modules' documentation.
- Directory `/cargodocs` to include following markdowns:
    - `centralized.md`
    - `decentralized.md`
    - `intro.md`
    - `lottery.md`
    - `overview.md`
    - `wdecentralized.md`
- `katex-header.html`, especially for LaTeX math symbols.

## Pre-submit checklist

- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
- PR
    - [ ] No clippy warnings in the CI
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reviewer requested
- Documentation

## Comments
- The existing `README` has been updated but will be changed and improved.
- It is based on the branch `raph@alba-bound`, so it will only include the centralized case documentation.
- For now, the content of `overview.md`, `intro.md`, and `centralized.md` are taken from the _hackmd_ document but will be revised.
- There will be cross-references between the code documentation and the markdown files.

Run the command to compile documentation locally:
```shell
RUSTDOCFLAGS="--html-in-header src/katex-header.html" cargo doc --no-deps --open
```

## Issue(s)

Relates to https://github.com/cardano-scaling/alba/issues/51


README.md
```diff
@@ -1,12 +1,33 @@
 # ALBA
+This is the rust library of the *Telescope* protocol proposed in the paper [Approximate Lower Bound Arguments](https://iohk.io/en/research/library/papers/approximate-lower-bound-arguments/), May 2024, Eurocrypt'24 by _Pyrros Chaidos, Prof Aggelos Kiayias, Leonid Reyzin, Anatoliy Zinovyev_.
 
-Implementation of _Approximate Lower Bound Arguments_ from the [paper](https://iohk.io/en/research/library/papers/approximate-lower-bound-arguments/) published by IOG Research:
+Alba is a generic protocol to prove succinctly a lower bound of the size of a *potentially* weighted set.
+Say we have a set $S_p$ of size $|S_p| \geq n_p$, and a lower bound $n_f < n_p$ of it we want to prove.
```
djetchev, line 5: We should clarify to the reader why we don't simply use `n_p = |S_p|`, i.e. ,why not use two parameters instead of three.

README.md
```diff
@@ -1,12 +1,33 @@
...
+Alba gives us a method to generate a proof of knowledge of this bound by finding the smallest subset of $S_p$ of size $u$ to convince a verifier.
```
djetchev, line 6: finding `the smallest` vs. `a smallest` - if we use a definite article this assumes that the set is unique. Also, is `u` really the cardinality of a smallest subset - in other words, can't the algorithm find a set that has a proper subset that will also convince a verifier. We should be precise here.

README.md
```diff
@@ -1,12 +1,33 @@
...
+The paper presents several schemes and optimizations.
+The basic scheme is enhanced in the "prehashed" version thanks to sorting $S_p$ with a *balls and bins* sorting algorithm reducing the number of hashes done per round.
```
djetchev, line 8: I would say : 

The basic scheme is extended to a `prehashed` variant by sorting $S_p$ with a *balls-and-bins* sorting algorithm, thus, reducing the number of hashes computed per round.

README.md
```diff
@@ -1,12 +1,33 @@
...
+A lottery scheme is also introduced to support Alba in a decentralized settings as well as a modification to use PRF in the CRS settings instead of using the ROM.
```
djetchev, line 9: We can briefly mention what exactly we mean by `decentralized setting`

cargodocs/centralized.md
```diff
@@ -0,0 +1,110 @@
+Telescope construction with prehashing and bounded DFS scheme given in _Section 3.2.2_.
+
+
+In some cases, $n_p$ may be small, which can affect the completeness and efficiency of the proof construction. The protocol adapts to such cases by introducing optimizations and tighter bounds.
```
djetchev, line 4: @tolikzinovyev :  If we are to give a name of `n_p` in English, what would it be? I still think of it as the cardinality of `S_p` and don't understand why it is a separate parameter.

cargodocs/centralized.md
```diff
@@ -0,0 +1,110 @@
...
+- In the basic ALBA scheme, optimal performance requires $n_p \geq \lambda^3$.
+- When $n_p$ is small (e.g., $n_p < \lambda^3$), the probability of successfully finding valid proof paths diminishes.
+- Ensure that the proof remains efficient and sound even when $n_p$ is small by introducing an improved protocol with tighter guarantees on the number of attempts and running time.
```
djetchev, line 7: what exactly do we mean (mathematically) by `n_p` being small in this context?

cargodocs/centralized.md
```diff
@@ -0,0 +1,110 @@
...
+- When $n_p$ is too small to guarantee a valid proof in a single attempt, the protocol allows the prover to make multiple attempts.
+- Each attempt is indexed by an integer $v \in \{1, \ldots, r\}$, where $r$ is the number of repetitions. This ensures higher completeness by amplifying the success probability.
+  - Set $r := \lambda_{\text{rel}}$, ensuring the completeness error becomes $2^{-\lambda_{\text{rel}}}$.
+  - This approach maintains completeness even when the prover has a limited number of elements in their set.
+
+### Optimized Proof Construction
+- A valid proof now consists of a tuple $(v, t, s_1, \ldots, s_u)$, where:
+  - $v$ is the repetition index.
+  - $t \in \{1, \ldots, d\}$ is another index used to explore proof paths.
```
djetchev, line 16: we should say what `d` is - also, what is a good way of explaining what `exploring proof paths` mean?

cargodocs/centralized.md
```diff
@@ -0,0 +1,110 @@
...
+- **Number of Repetitions:**  $r := \lambda_{\text{rel}}$ ensures completeness with error $2^{-\lambda_{\text{rel}}}$.
+- **Expected Running Time:** The average running time for the prover is: $n_p + \mathcal{O}(u^2)$.
+- **Worst-Case Running Time:** The worst-case running time is bounded by: $n_p + \mathcal{O}(u^2 \cdot \lambda_{\text{rel}})$.
+
+## Algorithm: Main Construction
+
+### Parameters
+- Soundness security parameter: $\lambda_{sec}$
+- Completeness security parameter: $\lambda_{rel}$
+- Approximate size of set $S_p$ to lower bound: $n_p$
```
djetchev, line 37: @tolikzinovyev : I suppose this is the parameter that causes most of the confusion for anyone who tries to apply Alba. I do understand the motivation from a complexity theory perspective (the original Babai's work - language recognition, approx upper/lower bound problem cannot be stated as a language recognition problem, etc.) but is going that deep in the complexity theory helping the user of Alba (plus, these concepts of Game of Nature, etc. are highly non-obvious to a non-expert on complexity theory)? For the purpose of the parameters, I'd advocate, as much as possible, to simplify the setting.

djetchev: In short, what exactly do I gain by having this parameter `n_p` as opposed to just working with `S_p`?

cargodocs/centralized.md
tolikzinovyev: In the paper we simply defined n_p = |S_p|. It's possible to consider a more general case: |S_p| >= n_p. Then we can maybe call n_p the required prover set size.

cargodocs/centralized.md
tolikzinovyev: As I said in another thread, in the paper we simply have n_p = |S_p|. We can consider |S_p| >= n_p since having more elements might in practice speed up the prover, but we need to be careful not to have |S_p| much larger than n_p because that would actually slow down the prover (the balls to bin hashing becomes the bottleneck). Also in the paper we assumed |S_p| = n_p when analyzing the prover running time. The DFS seems to terminate earlier in practice when |S_p| > n_p and it seems intuitive but we don't have a proof that it does not terminate later. I don't know how strict we want to be here, it's probably ok to just require n_p <= |S_p| <= 2 * n_p.

rrtoledo:
Perhaps we should enable the CI docs for this branch?

tolikzinovyev:
> Perhaps we should enable the CI docs for this branch?

CI is a good idea. Perhaps we could have a small preview of what documentation looks like before our code is more stable? Suggest making a PR with a skeleton: some placeholder text for the future documentation, enable CI, and move the website folder? We also should think about where the design doc will go.

PR #41, opened 2024-10-28
CI: treat warnings as errors.

tolikzinovyev:
## Content

Treat build warnings as errors. Additionally, build all targets (tests / benchmark / etc) and features (currently we don't use this).


djetchev (approved):

rrtoledo (approved):
lgtm

PR #50, opened 2024-11-01
Restructuring the code

rrtoledo:
## Content
Most of the code is inside a single file centralized_telescope.rs. This PR intends to split it in different files within a new folder centralized_telescope.

## Pre-submit checklist

- Branch
    - [ ] Tests are provided (if possible)
    - [ ] Commit sequence broadly makes sense
    - [ ] Key commits have useful messages
- PR
    - [ ] No clippy warnings in the CI
    - [ ] Self-reviewed the diff
    - [ ] Useful pull request description
    - [ ] Reviewer requested
- Documentation
    - [ ] Update README file (if relevant)
    - [ ] Update documentation website (if relevant)

## Comments

I have also created a new utils folder to put the shared types for all Alba algorithms, i.e. `Element` so far. 

## Issue(s)

Relates to #46

Issue #38, opened 2024-10-24
Better variable naming


rrtoledo:
## Why
For people not knowing the Alba paper well, the variable names are hard to follow. 

## What
We suggest here to change the variable names to be more descriptive and add a mapping to the paper.

rrtoledo:
Ideas from @curiecrypt:
```
Sp: prover_set
np: set_size
nf: threshold or verifier_threshold (?)
u: proof_size
d: search_depth
q: selection_probability
r: num_of_repetitions
b: bound
s: set_item
s_list: item_list
t: tuple_index
v: attempt_index
H0: item_hash
H1: prefix_hash
H2: final_hash
```

rrtoledo:
Ideas from @tolikzinovyev :
```
np - prover set size
nf - adversary set size
d - search width / number of search trees
q - final selection probability
r - number of tries (just to make it shorter, attempts is also fine)
b - DFS bound
s - element (to be consistent with the paper)
s_list - I'm guessing it's a (partial) tuple? tuple / element sequence
t - tree index
v - try index (just to make it short, attempt is also fine)
H0 - bin hash function
```

curiecrypt:
@rrtoledo, @djetchev, @tolikzinovyev. Here is a list of all suggestions we have. State your choice by putting `x` beside it.
Also, add more lines if you have more suggestions.

**Sp:**
- prover_set `x`
---
**np:** 
- set_size `x`
- prover set size
---
**nf:** 
- threshold `x`
- verifier_threshold 
- adversary set size
---
**u:** 
- proof_size `x`

**d:** 
- search_depth `x`
- search width 
- number of search trees
---
**q:** 
- selection_probability
- final selection probability `x`
---
**r:** 
- num_of_repetitions
- number of tries `x`
- attempts
---
**b:** 
- bound
- DFS bound `x`
---
**s:** 
- set_item
- element `x`
---
**s_list:** 
- item_list
- tuple
- element sequence `x`
---
**t:** 
- tuple_index 
- tree index `x`
---
**v:** 
- attempt_index `x`
- try index
---
**H0:** 
- item_hash
- bin hash function
- element_hash `x`
---
**H1:** 
- prefix_hash `x`
---
**H2:** 
- final_hash `x`

rrtoledo:
@curiecrypt I reordered the variables (to link `t` with `d`, and `v` with `r`), added a few options and voted.
[...]

PR #53, opened 2024-11-05
Updating variable names

rrtoledo:
## Content

Update variable names as discussed in https://github.com/cardano-scaling/alba/issues/38

## Comments

I changed `round_hash` (resp. `round_id`) to `hash` (resp. `id`) directly to dismiss clippy warnings.
As they are always used as `round.hash/id`, I thought this change was reasonable.

PR blocked by https://github.com/cardano-scaling/alba/pull/50
 
## Issue(s)

Relates to #38

### DIFF
--- a/src/centralized_telescope/init.rs
+++ b/src/centralized_telescope/init.rs
@@ -7,69 +7,75 @@ use std::f64::consts::LOG2_E;
 
 /// Setup algorithm taking a Params as input and returning setup parameters (u,d,q)
 pub fn make_setup(params: &Params) -> Setup {
-    let n_p_f64 = params.n_p as f64;
-    let n_f_f64 = params.n_f as f64;
+    let set_size_f64 = params.set_size as f64;
+    let lower_bound_f64 = params.lower_bound as f64;
 
-    let u_f64 = {
-        let numerator = params.lambda_sec + params.lambda_rel.log2() + 5.0 - LOG2_E.log2();
-        let denominator = (n_p_f64 / n_f_f64).log2();
+    let proof_size_f64 = {
+        let numerator =
+            params.soundness_param + params.completeness_param.log2() + 5.0 - LOG2_E.log2();
+        let denominator = (set_size_f64 / lower_bound_f64).log2();
         (numerator / denominator).ceil()
     };
 
-    let ratio = 9.0 * n_p_f64 * LOG2_E / ((17.0 * u_f64).powi(2));
+    let ratio = 9.0 * set_size_f64 * LOG2_E / ((17.0 * proof_size_f64).powi(2));
     let s1 = ratio - 7.0;
     let s2 = ratio - 2.0;
 
     if s1 < 1.0 || s2 < 1.0 {
-        // Small case, i.e. n_p <= λ^2
-        param_small_case(params, u_f64)
+        // Small case, i.e. set_size <= λ^2
+        param_small_case(params, proof_size_f64)
     } else {
-        let lambda_rel2 = params.lambda_rel.min(s2);
-        if u_f64 < lambda_rel2 {
-            // Case 3, Theorem 14, i.e.  n_p >= λ^3
-            param_high_case(params, u_f64, lambda_rel2)
+        let completeness_param2 = params.completeness_param.min(s2);
+        if proof_size_f64 < completeness_param2 {
+            // Case 3, Theorem 14, i.e.  set_size >= λ^3
+            param_high_case(params, proof_size_f64, completeness_param2)
         } else {
-            // Case 2, Theorem 13, i.e. λ^3 > n_p > λ^2
-            param_mid_case(params, u_f64, s1)
+            // Case 2, Theorem 13, i.e. λ^3 > set_size > λ^2
+            param_mid_case(params, proof_size_f64, s1)
         }
     }
 }
 
-fn param_small_case(params: &Params, u_f64: f64) -> Setup {
+fn param_small_case(params: &Params, proof_size_f64: f64) -> Setup {
     let ln12 = (12f64).ln();
-    let d = (32.0 * ln12 * u_f64).ceil();
+    let search_width = (32.0 * ln12 * proof_size_f64).ceil();
     Setup {
-        n_p: params.n_p,
-        u: u_f64 as u64,
-        r: params.lambda_rel as u64,
-        d: d as u64,
-        q: 2.0 * ln12 / d,
-        b: (8.0 * (u_f64 + 1.0) * d / ln12).floor() as u64,
+        set_size: params.set_size,
+        proof_size: proof_size_f64 as u64,
+        max_retries: params.completeness_param as u64,
+        search_width: search_width as u64,
+        leaf_probability: 2.0 * ln12 / search_width,
+        dfs_bound: (8.0 * (proof_size_f64 + 1.0) * search_width / ln12).floor() as u64,
     }
 }
 
-fn param_high_case(params: &Params, u_f64: f64, lambda_rel2: f64) -> Setup {
-    let l2 = lambda_rel2 + 2.0;
-    let d = (16.0 * u_f64 * l2 / LOG2_E).ceil();
-    debug_assert!(params.n_p as f64 >= d * d * LOG2_E / (9.0 * l2));
+fn param_high_case(params: &Params, proof_size_f64: f64, completeness_param2: f64) -> Setup {
+    let l2 = completeness_param2 + 2.0;
+    let search_width = (16.0 * proof_size_f64 * l2 / LOG2_E).ceil();
+    debug_assert!(params.set_size as f64 >= search_width * search_width * LOG2_E / (9.0 * l2));
     Setup {
-        n_p: params.n_p,
-        u: u_f64 as u64,
-        r: (params.lambda_rel / lambda_rel2).ceil() as u64,
-        d: d as u64,
-        q: 2.0 * l2 / (d * LOG2_E),
-        b: (((l2 + u_f64.log2()) / l2) * (3.0 * u_f64 * d / 4.0) + d + u_f64).floor() as u64,
+        set_size: params.set_size,
+        proof_size: proof_size_f64 as u64,
+        max_retries: (params.completeness_param / completeness_param2).ceil() as u64,
+        search_width: search_width as u64,
+        leaf_probability: 2.0 * l2 / (search_width * LOG2_E),
+        dfs_bound: (((l2 + proof_size_f64.log2()) / l2)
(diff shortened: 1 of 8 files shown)


src/centralized_telescope/setup.rs
```diff
@@ -0,0 +1,18 @@
+//! ALBA's Setup structure comprising, among others, the number of elements and internal parameters
+
+/// Setup output parameters
```
djetchev, line 3: Am I confused here, but I don't see the verifier threshold (the `n_f` in the paper)? Shouldn't it be part of the Setup struct?

src/centralized_telescope/setup.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct Setup {
+    /// Approximate size of prover set to lower bound
```
djetchev, line 6: I'd rename this : `Total number of elements available to the prover.` The term `approximate size` may confuse the reader.

src/centralized_telescope/setup.rs
rrtoledo: The verifier threshold (the n_f in the paper) is only needed for the setup parameter generations, hence it is not included in `Setup`

src/centralized_telescope/setup.rs
rrtoledo: I have an issue with "total" (as `|Sp| >= np` only), but if you are mitigating it with "available". I'll update it now.

src/centralized_telescope/setup.rs
```diff
@@ -0,0 +1,18 @@
...
+    pub set_size: u64,
+    /// Number of prover set's elements
+    pub proof_size: u64,
+    /// Maximum number of retries to find a proof
+    pub max_retries: u64,
+    /// Maximum number of subtrees to search to find a proof
+    pub search_width: u64,
+    /// Probability that a leaf was correctly chosen
```
djetchev, line 14: Now that I am thinking more and more about this name, we might be able to improve it even more as I'd be a bit confused about the term `leaf probability` if I am a first time reader. 

What is q: it is the probability that a random tuple `(t, s_1, \dots, s)u)` is a valid proof according to `h_2`. Maybe this can be said explicitly in the description to avoid confusion. Alternatively, it is the hardness. It sounds a bit more similar to `valid_proof_probability`?

src/centralized_telescope/setup.rs
```diff
@@ -0,0 +1,18 @@
+//! ALBA's Setup structure comprising, among others, the number of elements and internal parameters
+
+/// Setup output parameters
+#[derive(Debug, Clone, Copy)]
+pub struct Setup {
+    /// Approximate size of prover set to lower bound
+    pub set_size: u64,
+    /// Number of prover set's elements
```
djetchev, line 8: I'd change the comment to `The number of elements in a proof` to avoid confusion with `the total number of elements available to the prover at the proof generation`

src/utils/sample.rs
```diff
@@ -7,16 +7,16 @@ use std::cmp::min;
 pub(crate) fn sample_uniform(hash: &[u8], n: u64) -> Option<u64> {
     // Computes the integer reprensation of hash* modulo n when n is not a
     // power of two. *(up to 8 bytes, in little endian)
-    fn mod_non_power_of_2(hash: &[u8], n: u64) -> Option<u64> {
+    fn mod_noset_sizeower_of_2(hash: &[u8], n: u64) -> Option<u64> {
```
djetchev, line 10: confused with the function renaming

src/utils/sample.rs
```diff
@@ -7,16 +7,16 @@ use std::cmp::min;
...
         fn log_base2(x: u64) -> u64 {
             u64::from(
                 u64::BITS
                     .saturating_sub(x.leading_zeros())
                     .saturating_sub(1),
             )
         }
-        let epsilon_fail: u64 = 1 << 40; // TODO: update
-        let k = log_base2(n.saturating_mul(epsilon_fail));
+        let epsilolower_boundail: u64 = 1 << 40; // TODO: update
```
djetchev, line 18: what do we mean by this name ?

src/utils/sample.rs
```diff
@@ -7,16 +7,16 @@ use std::cmp::min;
...
+        let k = log_base2(n.saturating_mul(epsilolower_boundail));
         let k_prime: u64 = 1 << k;
```
djetchev, line 20: A bit puzzled about the variable name `k_prime` here considering we mean 2^k?

src/utils/sample.rs
```diff
@@ -38,22 +38,22 @@ pub(crate) fn sample_uniform(hash: &[u8], n: u64) -> Option<u64> {
...
 
-/// Takes as input a hash and probability $q$ and returns true with
+/// Takes as input a hash and probability q and returns true with
 /// probability q otherwise false according to a Bernoulli distribution
 /// (c.f. Appendix B, Alba paper).
 pub(crate) fn sample_bernoulli(hash: &[u8], q: f64) -> bool {
     // For error parameter ɛ̝, find an approximation x/y of q with (x,y) in N²
     // such that 0 < q - x/y <= ɛ̝
-    let epsilon_fail: u64 = 1 << 40; // TOOD: update
+    let epsilolower_boundail: u64 = 1 << 40; // TOOD: update
```
djetchev, line 51: idem - variable naming

src/centralized_telescope/proof.rs
```diff
@@ -0,0 +1,14 @@
...
+
+/// Alba proof
+#[derive(Debug, Clone)]
+pub struct Proof {
+    /// Numbers of retries done to find the proof
+    pub retry_counter: u64,
+    /// Index of the searched subtree to find the proof
+    pub search_counter: u64,
+    /// Sequence of elements from prover set
+    pub element_sequence: Vec<Element>,
```
djetchev, line 13: this vector apparently contains the proof. Should we give it a more descriptive name such as `proof_sequence`, `proof_elements` or `proof_content`?

src/centralized_telescope/init.rs
```diff
@@ -0,0 +1,114 @@
...
+            // Case 3, Theorem 14, i.e.  set_size >= λ^3
+            param_high_case(params, proof_size_f64, completeness_param2)
+        } else {
+            // Case 2, Theorem 13, i.e. λ^3 > set_size > λ^2
+            param_mid_case(params, proof_size_f64, s1)
+        }
+    }
+}
+
+fn param_small_case(params: &Params, proof_size_f64: f64) -> Setup {
```
djetchev, line 39: I'd provide a comment to explain what `small`, `mid` and `high` proof means? If possible, we can even try to give more descriptive names but at least an explanation in the code is important.

src/centralized_telescope/proof.rs
rrtoledo: The proof not only comprises a sequence of elements of the prover set but also the counters, as such `proof_content` might be misleading.

Each of these parameters are within a proof, and will be accessed as `proof.search/retry_counter" and `proof.element_sequence`. Repeating "proof" in the variable name seems a bit redundant in that aspect.

The reason why @curiecrypt and I liked the word "sequence" is because of the implied notion of order that you wouldn't have in "set" or "list".

src/utils/sample.rs
rrtoledo: This is a oversight of mine when changing all `n_f` to `lower_bound`. This variable is `epsilon_fail`.

src/utils/sample.rs
rrtoledo: This is a oversight of mine when changing all `n_p` to `set_size`.

src/centralized_telescope/init.rs
rrtoledo: There should be comments in the code above in `make_setup` (c.f. written below). Do you think they are not enough?
```
    if s1 < 1.0 || s2 < 1.0 {
        // Small case, i.e. set_size <= λ^2
        param_small_case(params, proof_size_f64)
    } else {
        let completeness_param2 = params.completeness_param.min(s2);
        if proof_size_f64 < completeness_param2 {
            // Case 3, Theorem 14, i.e.  set_size >= λ^3
            param_high_case(params, proof_size_f64, completeness_param2)
        } else {
            // Case 2, Theorem 13, i.e. λ^3 > set_size > λ^2
            param_mid_case(params, proof_size_f64, s1)
        }
    }
```

src/utils/sample.rs
rrtoledo: The name `k'` is not valid in rust so it was written in plain english. I'll update it to `new_n` (as it is used as `n` in `mod_power_2`)

src/centralized_telescope/setup.rs
curiecrypt: I think, giving an explicit description would solve the confusion. Since we have a dfs implementation, the name leaf probability seems appropriate but needs an explicit description of course.

src/utils/sample.rs
curiecrypt: +1

src/centralized_telescope/init.rs
curiecrypt: We would give descriptive documentation about the parameter cases. @rrtoledo let's open an issue for that

curiecrypt (dismissed):
LGTM 👍 
We should open issues for the topics that need to be documented.

src/centralized_telescope/setup.rs
rrtoledo: We chose `leaf_probability` as we really had in mind the trees of tuples and named several variables and functions accordingly (`H1 -> node_hash`, `H2 -> leaf_hash`, `d -> search_width` for the search arity).
You can see the random tuple as a candidate proof that you have to check, we considered it as the lead linked to a round that we were checking.

I'd suggest that we either add more descriptive comments here to explain succinctly what we are doing with each structure or change all the tree related names for consistency (`H1 -> round_hash`, `H2 -> proof_hash`, `d -> ?`).

src/centralized_telescope/init.rs
rrtoledo: @curiecrypt  If it is just adding a few comments, we can do it here no? I have already added quite a few in `prove_index` and so on afterall.

djetchev (dismissed):
Good work on this PR! 

LGTM

src/utils/sample.rs
rrtoledo: That was updated to `new_n`

src/centralized_telescope/setup.rs
rrtoledo: Updated

PR #61, opened 2024-11-08
Removing haskell code

rrtoledo:
## Content

Haskell code is not maintained and contains errors. It is the simplest to remove it until we write rust to haskell bindings


djetchev (approved):
lgtm

PR #63, opened 2024-11-12
Implement parameter derivation for simple lottery.

tolikzinovyev:
## Content

Parameter derivation for simple lottery  based on section 4.1 of the paper. Uses binary search to find the optimal value of `rs`.

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
- Documentation

## Issue(s)

### DIFF
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -56,7 +56,7 @@ mod tests {
         let setup = make_setup(&params);
         assert_eq!(1488, setup.proof_size);
         let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 2062.805);
+        assert!(mu > 2062.806);
         assert!(mu < 2062.807);
     }
 
@@ -71,8 +71,8 @@ mod tests {
         let setup = make_setup(&params);
         assert_eq!(1127, setup.proof_size);
         let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 1473.0);
-        assert!(mu < 1474.0);
+        assert!(mu > 1473.574);
+        assert!(mu < 1473.575);
     }
 
     #[test]
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -17,15 +17,15 @@ pub fn make_setup(params: &Params) -> Setup {
 
     loop {
         let middle = (left + right) / 2.0;
-        let rs = middle;
-        let rc = ratio / rs;
+        let ratio_soundness = middle; // rs
+        let ratio_completeness = ratio / ratio_soundness; // rc
 
-        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
-        let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
+        let lhs = params.soundness_param / (ratio_soundness.ln() - 1.0 + 1.0 / ratio_soundness);
+        let rhs = params.completeness_param / (ratio_completeness - 1.0 - ratio_completeness.ln());
 
         if (middle <= left) || (middle >= right) {
             let u = (lhs.max(rhs) * LN_2).ceil();
-            let mu = u * rc;
+            let mu = u * ratio_completeness;
             return Setup {
                 proof_size: u as u64,
                 lottery_probability: mu / params.set_size as f64,
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -9,24 +9,22 @@ use std::f64::consts::LN_2;
 pub fn make_setup(params: &Params) -> Setup {
     let ratio = params.set_size as f64 / params.lower_bound as f64;
 
-    // Execute binary search for the optimal value of `rs`.
+    // Execute binary search for the optimal value of `rs` in the interval
+    // ]1.0, ratio[.
 
-    let mut l: f64 = 0.0;
-    let mut r: f64 = ratio;
+    let mut left: f64 = 1.0;
+    let mut right: f64 = ratio;
 
     loop {
-        let m = (l + r) / 2.0;
-        let rs = m;
+        let middle = (left + right) / 2.0;
+        let rs = middle;
         let rc = ratio / rs;
 
         let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
         let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
 
-        if (m <= l) || (m >= r) {
-            let lhs = lhs * LN_2;
-            let rhs = rhs * LN_2;
-            let max = if lhs > rhs { lhs } else { rhs };
-            let u = max.ceil();
+        if (middle <= left) || (middle >= right) {
+            let u = (lhs.max(rhs) * LN_2).ceil();
             let mu = u * rc;
             return Setup {
                 proof_size: u as u64,
@@ -35,9 +33,9 @@ pub fn make_setup(params: &Params) -> Setup {
         }
 
         if lhs > rhs {
-            l = m;
+            left = middle;
         } else {
-            r = m;
+            right = middle;
         }
     }
 }
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -44,79 +44,90 @@ pub fn make_setup(params: &Params) -> Setup {
 mod tests {
     use super::super::params::Params;
     use super::make_setup;
+    use test_case::test_case;
 
-    #[test]
-    fn basic() {
-        let params = Params {
+    #[derive(Debug, Clone, Copy)]
+    struct Expected {
+        u: u64,     // proof size
+        mu_lb: f64, // average communication lower bound
+        mu_ub: f64, // average communication upper bound
+    }
+
+    #[test_case(
+        Params {
             soundness_param: 128.0,
             completeness_param: 128.0,
             set_size: 200,
             lower_bound: 100,
+        },
+        Expected {
+            u: 1488,
+            mu_lb: 2062.806,
+            mu_ub: 2062.807,
         };
-        let setup = make_setup(&params);
-        assert_eq!(1488, setup.proof_size);
-        let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 2062.806);
-        assert!(mu < 2062.807);
-    }
-
-    #[test]
-    fn different_lambdas() {
-        let params = Params {
+        "basic"
+    )]
+    #[test_case(
+        Params {
             soundness_param: 128.0,
             completeness_param: 64.0,
             set_size: 200,
             lower_bound: 100,
+        },
+        Expected {
+            u: 1127,
+            mu_lb: 1473.574,
+            mu_ub: 1473.575,
         };
-        let setup = make_setup(&params);
-        assert_eq!(1127, setup.proof_size);
-        let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 1473.574);
-        assert!(mu < 1473.575);
-    }
-
-    #[test]
-    fn fractional_ratio() {
-        let params = Params {
+        "different_lambdas"
+    )]
+    #[test_case(
+        Params {
             soundness_param: 128.0,
             completeness_param: 128.0,
             set_size: 150,
             lower_bound: 100,
+        },
+        Expected {
+            u: 4328,
+            mu_lb: 5264.558,
+            mu_ub: 5264.559,
         };
-        let setup = make_setup(&params);
-        assert_eq!(4328, setup.proof_size);
-        let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 5264.558);
-        assert!(mu < 5264.559);
-    }
-
-    #[test]
-    fn small_lambda_reliability() {
-        let params = Params {
+        "fractional_ratio"
+    )]
+    #[test_case(
+        Params {
             soundness_param: 128.0,
             completeness_param: 1.0,
             set_size: 200,
             lower_bound: 100,
+        },
+        Expected {
+            u: 527,
+            mu_lb: 554.495,
+            mu_ub: 554.496,
         };
-        let setup = make_setup(&params);
-        assert_eq!(527, setup.proof_size);
-        let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 554.495);
-        assert!(mu < 554.496);
-    }
-
-    #[test]
-    fn small_lambda_security() {
-        let params = Params {
+        "small_lambda_reliability"
+    )]
+    #[test_case(
+        Params {
             soundness_param: 1.0,
             completeness_param: 128.0,
             set_size: 200,
             lower_bound: 100,
+        },
+        Expected {
+            u: 358,
+            mu_lb: 672.362,
+            mu_ub: 672.363,
         };
+        "small_lambda_security"
+    )]
+    fn all(params: Params, expected: Expected) {
         let setup = make_setup(&params);
-        assert_eq!(358, setup.proof_size);
+        assert_eq!(expected.u, setup.proof_size);
         let mu = setup.lottery_probability * params.set_size as f64;
-        assert!(mu > 672.362);
-        assert!(mu < 672.363);
+        assert!(mu > expected.mu_lb);
+        assert!(mu < expected.mu_ub);
     }
 }
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -89,4 +89,34 @@ mod tests {
         assert!(mu > 5264.558);
         assert!(mu < 5264.559);
     }
+
+    #[test]
+    fn small_lambda_reliability() {
+        let params = Params {
+            soundness_param: 128.0,
+            completeness_param: 1.0,
+            set_size: 200,
+            lower_bound: 100,
+        };
+        let setup = make_setup(&params);
+        assert_eq!(527, setup.proof_size);
+        let mu = setup.lottery_probability * params.set_size as f64;
+        assert!(mu > 554.495);
+        assert!(mu < 554.496);
+    }
+
+    #[test]
+    fn small_lambda_security() {
+        let params = Params {
+            soundness_param: 1.0,
+            completeness_param: 128.0,
+            set_size: 200,
+            lower_bound: 100,
+        };
+        let setup = make_setup(&params);
+        assert_eq!(358, setup.proof_size);
+        let mu = setup.lottery_probability * params.set_size as f64;
+        assert!(mu > 672.362);
+        assert!(mu < 672.363);
+    }
 }
--- a/src/simple_lottery/init.rs
+++ b/src/simple_lottery/init.rs
@@ -7,6 +7,8 @@ use std::f64::consts::LN_2;
 
 /// Calculates Setup parameters.
 pub fn make_setup(params: &Params) -> Setup {
+    // The following follows section 4.1 of the ALBA paper. https://eprint.iacr.org/2023/1655
+
     let ratio = params.set_size as f64 / params.lower_bound as f64;
 
     // Execute binary search for the optimal value of `rs` in the interval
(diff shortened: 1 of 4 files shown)


src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
+//! Parameter derivation for simple lottery.
+
+use super::params::Params;
+use super::setup::Setup;
+
+use std::f64::consts::LN_2;
+
+/// Calculates Setup parameters.
+pub fn make_setup(params: &Params) -> Setup {
```
djetchev, line 9: Just a question: are we planning to expose this function to the public API (i.e., allow the users to call it)? If so, we might consider adding more specific comments on its usage - something like the `prove/verify` functions in @rrtoledo 's PRs. Here is a suggested template for the docs: 
```
 /// Adds two numbers together.
///
/// # Arguments
///
/// * `a` - The first integer to add.
/// * `b` - The second integer to add.
///
/// # Returns
///
/// An integer that is the sum of `a` and `b`.
///
/// # Examples
///
/// ```
/// let result = add(2, 3);
/// assert_eq!(result, 5);
/// ```
```
My comment is identical to the comments in @rrtoledo 's PRs.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+        let rs = m;
+        let rc = ratio / rs;
+
+        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
+        let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
+
+        if (m <= l) || (m >= r) {
+            let lhs = lhs * LN_2;
+            let rhs = rhs * LN_2;
+            let max = if lhs > rhs { lhs } else { rhs };
```
djetchev, line 28: Why not simply use 
```
let max = lhs.max(rhs);
```

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use std::f64::consts::LN_2;
+
+/// Calculates Setup parameters.
+pub fn make_setup(params: &Params) -> Setup {
+    let ratio = params.set_size as f64 / params.lower_bound as f64;
+
+    // Execute binary search for the optimal value of `rs`.
+
+    let mut l: f64 = 0.0;
```
djetchev, line 14: I know it's a pedantic personal preference :)  but I prefer giving slightly more descriptive names than `l`, `m`, `r`: 
`l` --> `left`
`r` --> `middle`
... 
specifically, I'll rename `rs` and `rc` to something more descriptive as it's not very intuitive what the abbreviations mean.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+        let rc = ratio / rs;
+
+        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
+        let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
+
+        if (m <= l) || (m >= r) {
+            let lhs = lhs * LN_2;
+            let rhs = rhs * LN_2;
+            let max = if lhs > rhs { lhs } else { rhs };
+            let u = max.ceil();
```
djetchev, line 29: e.g., we can replace `u` --> `proof_size`

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut r: f64 = ratio;
+
+    loop {
+        let m = (l + r) / 2.0;
+        let rs = m;
+        let rc = ratio / rs;
+
+        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
+        let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
+
```
djetchev, line 24: Since `left` and `right` are floating point numbers, don't you want to check some sort of convergence (e.g., if `right - left < EPS`, you break the loop)?

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::super::params::Params;
+    use super::make_setup;
+
+    #[test]
+    fn basic() {
```
rrtoledo, line 51: You should merge these 3 tests in one, say `test_correctness`, and iter on an array of `(Setup, expected_u, expected_mu)`, as the test names ("basic" and "fractional ratio") do not give much information and, so far, you have not distinguished in the parameter generation the case where the lambdas are equal.

rrtoledo (changes requested):
Overall, the code is clean and good. 
You have made an error on the dichotomic search lower bound and I think also the mu computation however. You could also process more quickly the case when the lambdas are equal.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[test]
+    fn different_lambdas() {
+        let params = Params {
+            soundness_param: 128.0,
+            completeness_param: 64.0,
+            set_size: 200,
+            lower_bound: 100,
+        };
+        let setup = make_setup(&params);
+        assert_eq!(1127, setup.proof_size);
```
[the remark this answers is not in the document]
curiecrypt: +1
I think we should be making assertions for some related bounds.
For example, if the params are set as above, then we should say that the calculated `u` and `mu` should be larger than some values like `limit_u` and `limit_mu`.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+        };
+        let setup = make_setup(&params);
+        assert_eq!(1488, setup.proof_size);
+        let mu = setup.lottery_probability * params.set_size as f64;
+        assert!(mu > 2062.805);
+        assert!(mu < 2062.807);
+    }
+
+    #[test]
+    fn different_lambdas() {
```
curiecrypt, line 66: Do we also need to test the following case:
```rust
let params = Params {
            soundness_param: 64.0,
            completeness_param: 128.0,
...
```

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Calculates Setup parameters.
+pub fn make_setup(params: &Params) -> Setup {
+    let ratio = params.set_size as f64 / params.lower_bound as f64;
+
+    // Execute binary search for the optimal value of `rs`.
+
+    let mut l: f64 = 0.0;
+    let mut r: f64 = ratio;
+
+    loop {
```
curiecrypt, line 17: This infinite loop construction concerns me a bit. If we insist on using this structure, we must ensure that we never have a case that lasts forever.
Even if we test it for "fair enough" number of times, can we say that we never get in to an infinite loop?

src/simple_lottery/init.rs
curiecrypt: +1
Also, we should adopt the style @rrtoledo used in [here](https://github.com/cardano-scaling/alba/blob/5c4b4c35c8565344fa4c01f6217c26ab11c5f9ed/src/simple_lottery.rs#L128)

curiecrypt (changes requested):
Thanks @tolikzinovyev, the algorithm seems simple and efficient. However, we should agree on some coding style.

src/simple_lottery/init.rs
tolikzinovyev: This is mostly a stylistic preference, but I prefer simple tests without loops since they give more confidence that they test the right thing (no bugs in the tests). Also, if a test fails, it is clear which one fails, unlike with a loop.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn basic() {
+        let params = Params {
+            soundness_param: 128.0,
+            completeness_param: 128.0,
+            set_size: 200,
+            lower_bound: 100,
+        };
+        let setup = make_setup(&params);
+        assert_eq!(1488, setup.proof_size);
+        let mu = setup.lottery_probability * params.set_size as f64;
```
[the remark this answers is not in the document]
tolikzinovyev: No, `mu` is just `p * n_p`, doesn't need to be integer.

src/simple_lottery/init.rs
tolikzinovyev: I don't think so. The point of this test is to check that the code uses the right lambdas in the right places. For example, a buggy implementation could use only one lambda, or swap the two lambdas.

src/simple_lottery/init.rs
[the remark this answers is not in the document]
tolikzinovyev: @rrtoledo calculated by hand
@curiecrypt I'm not sure what you are suggesting, what would `limit_u` and `limit_mu` be?

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let mut l: f64 = 0.0;
+    let mut r: f64 = ratio;
+
+    loop {
+        let m = (l + r) / 2.0;
+        let rs = m;
+        let rc = ratio / rs;
+
+        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
```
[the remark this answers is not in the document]
rrtoledo: `soundness_bound` then? Any variant of it to make it more descriptive and to make it more distinctive from the `l` and `r` afterwards would be great.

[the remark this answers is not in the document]
rrtoledo: Precomputing one variable shouldn't affect readability that much, especially if you name it well (e.g. `one_over_rs` or `rs_inv`).

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn make_setup(params: &Params) -> Setup {
+    let ratio = params.set_size as f64 / params.lower_bound as f64;
+
+    // Execute binary search for the optimal value of `rs`.
+
+    let mut l: f64 = 0.0;
+    let mut r: f64 = ratio;
+
+    loop {
+        let m = (l + r) / 2.0;
```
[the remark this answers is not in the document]
rrtoledo: This can be done with more comments?

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+use super::params::Params;
+use super::setup::Setup;
+
+use std::f64::consts::LN_2;
+
+/// Calculates Setup parameters.
+pub fn make_setup(params: &Params) -> Setup {
+    let ratio = params.set_size as f64 / params.lower_bound as f64;
+
+    // Execute binary search for the optimal value of `rs`.
```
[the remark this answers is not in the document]
rrtoledo: It is much more efficient though. 
Yes, we shouldn't run the setup many times, and it can be done beforehand, but I do not think adding a 6 line long block would make the code less readable or harder to maintain.

src/simple_lottery/init.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        let lhs = params.soundness_param / (rs.ln() - 1.0 + 1.0 / rs);
+        let rhs = params.completeness_param / (rc - 1.0 - rc.ln());
+
+        if (m <= l) || (m >= r) {
+            let lhs = lhs * LN_2;
+            let rhs = rhs * LN_2;
+            let max = if lhs > rhs { lhs } else { rhs };
+            let u = max.ceil();
+            let mu = u * rc;
```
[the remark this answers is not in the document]
rrtoledo: I do not say `u` is not an integer. I was reading the formula from the paper that uses the bound and not `u` directly and so hoped for a better value:
```µ ≥ max { λsec · ln 2 / (ln rs − 1 + 1/rs) · rc, λrel · ln 2 / (rc − 1 − ln rc) · rc }.```
Peraphs the paper is wrong and there should be some `ceil` here.

File: src/simple_lottery/init.rs (hunk 13 above)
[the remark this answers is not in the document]
Comment (rrtoledo) [reply src/simple_lottery/init.rs]: Perhaps French or western european? I am not sure. I think it is to differentite from tuples `(,)`.
p.s. In France, we would use  double brackets `]] , [[` to precise we are working with integer (of course this is not the case here).

File: src/simple_lottery/init.rs (hunk 10 above)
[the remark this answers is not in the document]
Comment (rrtoledo) [reply src/simple_lottery/init.rs:58]: That's not how it is defined in the paper though (p is defined by mu):
```
The simple lottery scheme is parametrized by the expected number of network participants µ. Let H be a random oracle that outputs 1 with probability p = µ / np and 0 otherwise.
```

File: src/simple_lottery/init.rs (hunk 6 above)
Comment (rrtoledo) [reply src/simple_lottery/init.rs:49]: There are no differences between the tests though (they test the exact same function, which has no forks/cases). It does not make sense to separate them. Especially if you want to add more values, it means writing another test instead of just adding a test vector.

File: src/simple_lottery/init.rs (hunk 5 above)
Comment (rrtoledo) [reply src/simple_lottery/init.rs:25]: I would not say that the goal of doing this is to optimize the code though but to follow the logic of what we want to do: we want to find `rs` such that the u given by each bounds are equal.

File: src/simple_lottery/init.rs (hunk 1 above)
Comment (rrtoledo) [reply src/simple_lottery/init.rs:9]: Already done in https://github.com/cardano-scaling/alba/pull/66

File: src/simple_lottery/init.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (rrtoledo) [reply src/simple_lottery/init.rs:72]: I think @curiecrypt just wants to check that the values you get are greater than the bounds, not equal to some values.

File: src/simple_lottery/init.rs (hunk 6 above)
Comment (tolikzinovyev) [reply src/simple_lottery/init.rs]: I changed the tests to use parameterized test instances which reduce code duplication. Other benefits:
1. no complicated logic in tests (loops)
2. each test case is treated as its own test, so if a test fails, you know which one
```running 5 tests
test simple_lottery::init::tests::all::basic ... ok
test simple_lottery::init::tests::all::different_lambdas ... ok
test simple_lottery::init::tests::all::small_lambda_reliability ... ok
test simple_lottery::init::tests::all::fractional_ratio ... ok
test simple_lottery::init::tests::all::small_lambda_security ... ok
```
Take a look!

src/simple_lottery/init.rs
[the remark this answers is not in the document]
tolikzinovyev: It would add an extra code path with no clear benefit. I'm against extra complexity until it is justified by real requirements.

src/simple_lottery/init.rs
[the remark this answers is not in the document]
tolikzinovyev: What is the advantage of your proposal? I'm currently unable see what it would improve.

src/simple_lottery/init.rs
tolikzinovyev: renamed `rs`, `rc` for consistency

src/simple_lottery/init.rs
rrtoledo: Let's go a step backward to see what we are trying to achieve. 

The code here has a clear goal: finding the minimal `u`. We do so by doing a dichotomic search on `rs`. What are the conditions when iterating on `rs`? We have 3: `rs` and `rc` must be stricter than  - this is already implicitely handled by the bounds of the dichotomic search- and we want to converge the soundness and completeness bounds, that is `|(lhs * ln2).ceil() - (rhs * ln2).ceil()| = 0` .

Now, this repository is supposed to be a reference implementation, you say yourself in the design document. Hence, we are aiming at being clear about what each piece of code is doing. Do we think that we should have undocumented code that goes straight for the solution without explaining some intricacies, not only about the algorithm but also how it was implemented (for instance the variable names, why looping as done here is fine and the expected number of iterations (your talk about f64 precision))? Or should we make an effort to be clear by having meaningful and coherent variable names, checks and tests? I lean towards the latter.

Most of the review comments, and more globally general decisions, so far have aimed at that goal. There has of course also been here and there some changes and optimizations asked, I won't deny that. But if you look at the history of this repository, you will see that we have already made some along the way that were deemed reasonable at some point in time (for instance, reusing values in the `centralized_telescope` setup) and so it is coherent to continue in that fashion. It is in the developer's, reviewer's best interest to keep a similar coding style afterall.

 To go back to the bound, `|(lhs * ln2).ceil() - (rhs * ln2).ceil()| = 0`. 
> Aren't I talking about optimizing the code here? 

I am, it is an optimization, and that is fine. It is also not my only point.
    
> Which of the goals does the current code not solve?

The code is not the clearest. We should strive for better variable names, more documentation, and  specifying what we are trying to do. The paper moreover can be hard to understand, you highlighted before some of its weaknesses and people have been confused about parts of it (the role of `nf` and `np` for instance as confounded more than one person, I also personally discussed with Pyrrhos more than once about part of the paper). We have an opportunity here to improve on it, let's do it.

src/simple_lottery/init.rs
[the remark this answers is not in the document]
rrtoledo: The name `lhs` only stands for "left hand side". Why left? Is it simply because it is written on the left of the bound check in the paper or for some other reason? I do not know.

At the moment, `lhs` is less descriptive and meaningful than `soundness_bound`. This value moreover is still a bound for `u,` perhaps loose, right, perhaps not sufficient but still necessary. If this is a problem, we can always multiply it by `ln2`.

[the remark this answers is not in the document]
rrtoledo: I have mostly mentioned this to mirror similar optimizations we have made on `centralized_telescope`'s setup. 

It might be a small optimization, but for coherency, it would be better to keep the same coding style.

src/simple_lottery/init.rs
rrtoledo: Adding more tests now is indeed easier, thanks a lot for that.

This change however does not change the fact that you are not testing different cases of the code, or checking explicitely that the lambdas are computed correctly. Hence the way you split the tests seem artificial (why checking in particular smalll lambda reliability if the formula is the same regardless on how high or low this lambda is?).

If you do want to check something precisely, I would encourage you to use loops contrary to what you say. Having only one test_vector is putting the code at risk of a false positive, the code could also be tailored for this specific value.

tolikzinovyev: > This change however does not change the fact that you are not testing different cases of the code

Can you elaborate what you mean? What cases do I not check?

> checking explicitely that the lambdas are computed correctly

lambdas are not computed, they are inputs, I'm not sure what you mean

> why checking in particular smalll lambda reliability

to test that the binary search moves in the right direction

> If you do want to check something precisely, I would encourage you to use loops

Could you explain what this loop would look like? What does iterate over and what does it check in each iteration? Pseudo-code would be ideal.

> the code could also be tailored for this specific value

This can be true of any code. Tests generally do not aim to test all possible inputs. The goal is only to test different "behaviors" of the code (including all edge cases).

src/simple_lottery/init.rs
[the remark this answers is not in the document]
tolikzinovyev: I don't know what change you are proposing though. Note that the expected value of a random variable is often fractional even when the random variable itself is. For example, the expected value of a normal 6-sided die after a roll is (1+2+...+6) / 6 = 3.5.

src/simple_lottery/init.rs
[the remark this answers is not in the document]
rrtoledo: There is a clear benefit. We would not have to go through a loop and directly output the result. We would also be closer to the paper, which always is helpful for a reference implementation.

src/simple_lottery/init.rs
tolikzinovyev: I like randomized tests but they never replace basic deterministic unit tests. In this case, what would you check in randomized tests? I still don't see what you are proposing exactly.

src/simple_lottery/init.rs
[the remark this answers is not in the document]
tolikzinovyev: > We would not have to go through a loop and directly output the result.

And what would be the benefit of that?

src/simple_lottery/init.rs
tolikzinovyev: What I mean is, mathematically `lhs` and `rhs` exist, but given that we use floating-point numbers with fixed precision, there might not a representation that satisfy the formula.

What comments would you like to add?

rrtoledo (approved):
LGTM

curiecrypt (approved):
LGTM

Issue #59, opened 2024-11-07
Restruct code (2)


rrtoledo:
## Why

Add single point of access to the user with a wrapper class to facilitate generics and avoid parameter redundancy

## What

![image](https://github.com/user-attachments/assets/b8ff2548-5042-4cac-87c3-4c641dd99490)

## Tasks
- [x] simple wrapper #73 
- [x] refactoring #66 
- [x] move tests #79 
- [ ] ? from/to

PR #66, opened 2024-11-12
Add wrapper structure to work on

rrtoledo:
## Content

Restructure the code to have a user facing structure from which we can directly prove and verify.

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
- Documentation

## Comments

I removed the Params structure as it was not really needed and renamed the Setup one "Params".
I moved back the code into impls so that it is easier to use generics.

## Issue(s)

Relates to #59


rrtoledo:
I used "telescope" as name for the wrapper structure, I am more than happy to change this.

rrtoledo:
@djetchev @curiecrypt Are the comments for public functions reasonable? They could be more verbose perhaps?

tolikzinovyev (changes requested):
I'm not a big fan of the interface of `Telescope::new()`. This will just encourage the user to use user-friendly parameters instead of hardcoding `Setup`, which goes against https://github.com/cardano-scaling/alba/blob/8d8bcbfb130c58d57e7b6148c74737c35babc835/design/design.md#parameter-calculation.

I think `new()` should accept `Setup` only.

rrtoledo:
> I'm not a big fan of the interface of `Telescope::new()`. This will just encourage the user to use user-friendly parameters instead of hardcoding `Setup`, which goes against https://github.com/cardano-scaling/alba/blob/8d8bcbfb130c58d57e7b6148c74737c35babc835/design/design.md#parameter-calculation.
> 
> I think `new()` should accept `Setup` only.

And I strongly disagree with you. We should not make giving hardcoded parameters the norm, as we have no insurance they were computed correctly, this should also go against usable security.

curiecrypt (dismissed):
I am OK with this wrapper structure. It seems professional, user-friendly, and easy to read.
I just have 2 minor comments.
LGTM 👍

curiecrypt:
> I used "telescope" as name for the wrapper structure, I am more than happy to change this.

I tried to put another name but couldn't find a better one. We might think of a more creative one in the future.

tolikzinovyev:
I see your point @curiecrypt. More importantly, I think, we need to be coherent as a team. A design decision should first be discussed on the design document _before_ there is a pull request that contradicts it. Otherwise, what is the point of the design document?

rrtoledo:
Replaced by https://github.com/cardano-scaling/alba/pull/116

PR #67, opened 2024-11-12
Design document

tolikzinovyev:
## Content

The first design document for our implementation. Superceeds #29.

PR #72, opened 2024-11-19
Move website deployment out of Rust CI

tolikzinovyev:
## Content

The website stuff shouldn't be part of Rust CI. Also, the skipped docs check that never runs on pull requests is a bit annoying.

Screenshot 2024-11-18 at 23-10-37 Design document by tolikzinovyev · Pull Request #67 · cardano-scaling_alba

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -1,0 +1,30 @@
+name: Deploy Website
+
+on:
+  push:
+    branches: ["main"]
+
+jobs:
+  deploy:
+    permissions:
+      contents: write
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: ✅ Checkout
+        uses: actions/checkout@v4
+
+      - name: 🪨 Setup mdBook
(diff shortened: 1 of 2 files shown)


rrtoledo:
Which issue is this PR linked to?

rrtoledo (changes requested):
While we should not care about deploying the website on the CI, we definitely should build cargo doc to make sure that the documentation can be built and the examples do not fail (that would be embarassing).
If it is not the case (I haven't checked), I would like to add it in this PR (as you already talk about skipped doc in the CI).

Also can you write an issue to attach to this PR?

curiecrypt (approved):
LGTM
After linking it to an issue, it can be merged. 
I will take care of building the cargo doc, issue #74. @rrtoledo

tolikzinovyev:
I'm not sure we need an issue for such small changes. We don't want to introduce friction for improving the codebase. What do you think?

rrtoledo (approved):

PR #111, opened 2024-12-16
Check elements are distinct in simple lottery.

tolikzinovyev:
## Content

Found another issue with the simple lottery. We need to check that elements in the proof are distinct. I guess the simple lottery is not so simple. :)

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
- Documentation

### DIFF
--- a/src/simple_lottery/algorithm.rs
+++ b/src/simple_lottery/algorithm.rs
@@ -8,6 +8,8 @@ use crate::utils::types::Element;
 use blake2::{Blake2s256, Digest};
 
 pub(super) fn prove(setup: &Setup, prover_set: &[Element]) -> Option<Proof> {
+    debug_assert!(crate::utils::misc::check_distinct(prover_set));
+
     let mut element_sequence = Vec::with_capacity(setup.proof_size as usize);
     for &element in prover_set {
         if lottery_hash(setup, element) {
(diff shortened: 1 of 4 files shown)


src/simple_lottery/algorithm.rs
```diff
@@ -14,6 +14,7 @@ pub(super) fn prove(setup: &Setup, prover_set: &[Element]) -> Option<Proof> {
             element_sequence.push(element);
         }
         if prover_set.len() as u64 >= setup.proof_size {
+            element_sequence.sort();
```
rrtoledo, line 17: These elements are not necessarily unique

rrtoledo (changes requested):
I think we need to add another check in the prover

src/simple_lottery/algorithm.rs
tolikzinovyev: I think we assume that `prover_set` must have distinct elements. Let's add this requirement to the function's doc? For now I added a debug assert.

rrtoledo: Sure, I'll add a warning in the doc PR for the lottery then!

rrtoledo (approved):
LGTM

curiecrypt (approved):
LGTM

rrtoledo (approved):
LGTM

PR #114, opened 2024-12-18
Example - Centralized Telescope with BLS Signatures

### LINKED ISSUE #78: Example - Threshold signature
## Why

This issue aims at using Centralized Telescope algo. in a example to showcase how to use it.

## What

The example to implement will be a threshold signature scheme where a centralized party shall provide a proof of lower bound of some vote.

## Tasks

TODO later
- [ ]

curiecrypt:
## Content
This PR introduces an example demonstrating the ALBA Threshold Signature scheme with the following workflow:

### Workflow
1. **Define Telescope Parameters**:
   - Set `nb_elements` to 1,000.
   - Configure `soundness_param` and `completeness_param` as 128.0.
   - Calculate the prover set size (`set_size`) as 80% of `nb_elements` and the lower bound as 20%.
   - Initialize the Telescope with these parameters.
2. **Initialize Signers**:
   - Create a list of candidate `Signer`s with randomly generated keys.
3. **Start Registration**:
   - Open a new `Registration`.
   - Select and register a subset of candidate signers, tracking the number of successfully registered signers.
4. **Close Registration**:
   - Finalize the registration by computing the checksum of registered keys.
   - Update registered signers with the computed checksum.
5. **Generate Individual Signatures**:
   - Select a subset of registered signers.
   - Generate `IndividualSignature`s for each based on the input message.
6. **Generate ALBA Threshold Signature**:
   - Aggregate valid individual signatures to create an `AlbaThresholdSignature`.
   - Print details of the generated threshold signature if successful.
7. **Verify ALBA Threshold Signature**:
   - Validate the generated threshold signature using the Telescope, registration, and input message.
   - Print whether the verification succeeded or failed.

### Purpose
This example provides a detailed implementation of the ALBA Threshold Signature workflow, showcasing the integration of Telescope parameters, signer registration, individual signature generation, and threshold signature verification.

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

## Issue(s)
Closes #78


examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub(crate) fn prove<const N: usize>(
+        params: &Params,
+        signatures: &[IndividualSignature],
+        set_size: u64,
+        closed_registration: &ClosedRegistration,
+        msg: &[u8],
+    ) -> Option<Self> {
+        let try_aggregate =
+            AggregateSignature::aggregate::<N>(signatures, closed_registration, msg, set_size);
+        if let Some(aggregate) = try_aggregate {
```
djetchev, line 45: In lines 45--59, you seem to be chaining multiple operations that return `Option`'s. Rust provides the `and_then()` operator to simplify that chaining. Here is how you code excerpt would look like with `and_then`: 
```
 // Combine the operations to avoid multiple unwrapping
 AggregateSignature::aggregate::<N>(signatures, closed_registration, msg, set_size)
    .and_then(|aggregate| {
        let prover_set = aggregate.create_prover_set::<N>();
        let alba = CentralizedTelescope::create(params);
        alba.prove(&prover_set).map(|proof| Self { aggregate, proof })
    })
```

examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
...
+        msg: &[u8],
+    ) -> bool {
+        let commitment: [u8; N] =
+            ClosedRegistration::get_commitment(&closed_registration.check_sum, msg);
+
+        if commitment != self.aggregate.commitment.as_slice() {
+            return false;
+        }
+
+        for sig in self.aggregate.valid_signatures.clone() {
```
djetchev, line 79: since you are not modifying the `valid_signatures` during iteration, it is more efficient and more idiomatic to use `iter()` instead of `clone()`, that is 
```
if !self.aggregate.valid_signatures
    .iter()
    .all(|sig| sig.verify::<N>(&commitment, closed_registration))
{
    return false;
}
```

examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
...
+        closed_registration: &ClosedRegistration,
+        msg: &[u8],
+    ) -> Option<Self> {
+        let try_aggregate =
+            AggregateSignature::aggregate::<N>(signatures, closed_registration, msg, set_size);
+        if let Some(aggregate) = try_aggregate {
+            let prover_set: Vec<Element> = aggregate.create_prover_set::<N>();
+            let alba = CentralizedTelescope::create(params);
+            let try_proof = alba.prove(&prover_set);
+            if let Some(proof) = try_proof {
```
rrtoledo, line 49: As you are directly consuming the proof you could do this `if let Some(proof) = alba.prove(&prover_set) {` directly

examples/centralized_telescope.rs
rrtoledo: Since all these signatures are BLS signatures you could batch verify them!

You could implement this in `aggregate.rs` using blst's function 
```
pub fn aggregate_verify(
                &self,
                sig_groupcheck: bool,
                msgs: &[&[u8]],
                dst: &[u8],
                pks: &[&PublicKey],
                pks_validate: bool,
            ) -> BLST_ERROR 
```

examples/threshold_signature/signature.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Individual Signature.
+/// It includes a blst signature, its verification key, and the registration index of the signer.
+#[derive(Debug, Clone)]
+pub(crate) struct IndividualSignature {
+    pub(crate) signature: Signature,
+    pub(crate) verification_key: PublicKey,
+    pub(crate) signer_index: usize,
+}
+
+impl IndividualSignature {
```
rrtoledo, line 16: Not sure if it makes sense but prove/verify could be done in your signer (what I renamed "RegisteredSigner" in a comment) impl?

examples/threshold_signature/registration.rs
```diff
@@ -0,0 +1,106 @@
+use blake2::digest::{Update, VariableOutput};
+use blake2::Blake2bVar;
+use blst::min_sig::PublicKey;
+use std::collections::HashMap;
+
+/// A hash table for storing BLS public keys with their indices
+#[derive(Debug, Clone)]
+pub(crate) struct RegisteredKeys {
```
rrtoledo, line 8: I am not sure you actually need the signer index, but only being able to know they are in the struct. Perhaps using a [hashset](https://doc.rust-lang.org/std/collections/hash_set/struct.HashSet.html) instead would be more efficient?

rrtoledo (changes requested):
Good work!

Some suggestions here and there to simplify the code (usnig type alias instead of struct, merging impl block potentially...) and make it more efficient (not using maps but sets, batching the verification).

I think the example would improve in clarity if you rename the folder "threshold_signature" "aggregate_signature" as we are only using these signature list (from registered signers) to create a threshold signature in centralized_telescope.rs. Perhaps renaming centralized_telescope.rs "centralized_threshold.rs" also?

tolikzinovyev:
Thanks for working on this @curiecrypt! Do you need my review or @rrtoledo and @djetchev will be enough?

examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
...
+        params: &Params,
+        signatures: &[IndividualSignature],
+        set_size: u64,
+        closed_registration: &ClosedRegistration,
+        msg: &[u8],
+    ) -> Option<Self> {
+        let try_aggregate =
+            AggregateSignature::aggregate::<N>(signatures, closed_registration, msg, set_size);
+        if let Some(aggregate) = try_aggregate {
+            let prover_set: Vec<Element> = aggregate.create_prover_set::<N>();
```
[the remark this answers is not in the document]
curiecrypt: This is controlled in `AggregateSignature::aggregate::<N>(signatures, closed_registration, msg, set_size);`.
Do we still need to check it here as well?

examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
...
+use crate::threshold_signature::signature::IndividualSignature;
+use crate::threshold_signature::signer::{Candidate, Signer};
+use threshold_signature::aggregate::AggregateSignature;
+
+const DATA_LENGTH: usize = 32;
+pub(crate) type Element = [u8; DATA_LENGTH];
+
+/// Alba proof with aggregate signature
+#[derive(Debug, Clone)]
+pub struct AlbaThresholdProof {
```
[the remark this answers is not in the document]
curiecrypt: Actually, I had a hard time while deciding this struct's name. 
I went with this because we have an Alba proof and an aggregate signature which create a threshold signature proof. 
I would have called it `AlbaThresholdSignatureProof`, but that is way too long, IMO.
WDYT?

examples/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: If it's already done, no! :)

examples/threshold_signature/signature.rs
```diff
@@ -0,0 +1,49 @@
...
+        &self,
+        commitment: &[u8],
+        closed_registration: &ClosedRegistration,
+    ) -> bool {
+        if closed_registration
+            .registered_keys
+            .is_key_at_index::<N>(&self.verification_key, self.signer_index)
+        {
+            let result =
+                self.signature
```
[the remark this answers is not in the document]
curiecrypt: I named `Hash(check_sum || msg)` as commitment.

examples/threshold_signature/signature.rs
```diff
@@ -0,0 +1,49 @@
...
+                    .verify(false, commitment, &[], &[], &self.verification_key, false);
+            return result == BLST_ERROR::BLST_SUCCESS;
+        };
+        false
+    }
+
+    /// Return the hash of the signature and its public key
+    /// This function is used to create the `prover_set` of Alba protocol.
+    pub fn to_element<const N: usize>(&self) -> [u8; N] {
```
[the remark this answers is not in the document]
curiecrypt: My intention was to use the byte representation of the signature as an element. Even, I changed the `DATA_LENGTH` to 48B. But, as we discussed previously, we can keep the `DATA_LENGTH` as it is by hashing the signatures.

examples/centralized_telescope.rs
```diff
@@ -0,0 +1,146 @@
+//! Example
+
+// REMOVE!!!!!!!!!!!!!!!!
+#![allow(dead_code)]
+
+mod threshold_signature;
```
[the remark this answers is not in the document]
rrtoledo: I think that by calling this example file `centralized_telescope.rs`, you want to highlight the fact that this is an example of use of this protocol and so prepare for additional examples in the future.

I suggested renaming it `centralized_threshold.rs` (or similar) so that readers have a better idea of what the example is (without opening the file), yet still linked to the centralized scheme.

I think both approach have merits. If you prefer keeping the name of the file as is, perhaps adding a md file in /example/ to succinctly describe each of the examples (right now just this one) might be helpful.

examples/centralized_telescope.rs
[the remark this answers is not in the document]
rrtoledo: I agree with you that `AlbaThresholdSignatureProof`. I am partial towards signature, as we are trying to implement a threshold signature using Alba, but I can see you are trying to emphasize Alba which is also fine. :)

examples/threshold_signature/signature.rs
rrtoledo: @curiecrypt any opinion? Feel free to mark the conversation as resolved :)

tolikzinovyev (commented):
I think this example is heavily inspired by Mithril which nice. The problem is, it is too advanced. I spent 30 minutes looking at it and I still don't have a good understanding. For people not familiar with alba, this would be even more difficult.

Some of the things that make it complicated are registration and commitments. I know it would be a big overhaul, but I think we could consider something simpler for the demo example. One idea is to implement a multisignature scheme defined in `alba-definitions.pdf` that I shared yesterday on slack, Definition 3. There is no registration or commitments there and we would just implement four functions: `KeyGen`, `Sign`, `Aggregate`, `Verify`. Such an implementation would be under 100 lines of code.

What do you think?

tolikzinovyev:
Also, I definitely support changing the element size to 48 bytes to fit the signature.

examples/aggregate_signature/helpers.rs
```diff
@@ -0,0 +1,102 @@
...
+    signature_list: &[IndividualSignature],
+    registration: &Registration,
+    msg: &[u8],
+) -> HashMap<Element, usize> {
+    let mut valid_signatures = HashMap::new();
+
+    for sig in signature_list {
+        if let Some(verification_key) = registration.registered_keys.get(&sig.index) {
+            let Some(checksum) = &registration.checksum else {
+                println!("Error: Registration is not closed. Cannot verify signatures.");
```
rrtoledo, line 35: You might print this lines many times, perhaps you could move it outside the loop?

rrtoledo: We might also not collect the signatures if the registration is not closed (Unless you want to show something in a future example)?

examples/aggregate_signature/helpers.rs
```diff
@@ -0,0 +1,102 @@
...
+    let mut signatures = Vec::with_capacity(alba_threshold_signature.proof.element_sequence.len());
+    for sig_bytes in &alba_threshold_signature.proof.element_sequence {
+        if let Ok(signature) = Signature::from_bytes(sig_bytes.as_slice()) {
+            signatures.push(signature);
+        } else {
+            println!("Error: Failed to parse signature from bytes.");
+            return false;
+        }
+    }
+    let signature_refs: Vec<&Signature> = signatures.iter().collect();
```
rrtoledo, line 66: Could you not push the refs in signatures to begin with?

examples/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Code structure
+### Signer
+- There are two types of signers:
+  - `Signer`: 
+    - Candidate signer, 
+    - including only the signing key and its verification key.
+    - Initialization of a new signer is done by `init` function.
+    - New registered signer can be created based on this signer with closed registration by `new_signer` function.
+  - `RegisteredSigner`:
```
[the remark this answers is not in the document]
rrtoledo: Or using an option*

examples/aggregate_signature/registration.rs
```diff
@@ -0,0 +1,83 @@
...
+            println!("Registration is closed!");
+            return false;
+        }
+
+        if self.registered_keys.values().any(|v| *v == key) {
+            println!("Key already registered!");
+            return false;
+        }
+
+        self.registered_keys.insert(index, key);
```
rrtoledo, line 38: You may be overwriting the key (if the index is the same for 2 parties)!

rrtoledo: I think it would be more natural for the register to output a Option<integer> where the integer is the signer's index (or none if already registered). 
You could instead check the length of the interator and return this as the index?

examples/aggregate_signature/signer.rs
```diff
@@ -0,0 +1,75 @@
...
+        let vk = sk.sk_to_pk();
+
+        Self {
+            signing_key: sk,
+            verification_key: vk,
+        }
+    }
+
+    /// Create a new registered signer if the signer's key exists in the closed registration
+    pub(crate) fn new_signer<const N: usize>(
```
rrtoledo, line 44: It might be worth having two functions here: one to register the signer on an acc (just calling the register of the acc) and a second to get the checksum.
Why so?
If we want to model registration systems, we would have each party register dynamically. Then the registration closes and everyone gets the checksum and proof of accumulation (e.g. Merkle path).

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Create Alba proof and extract indices of proof elements.
+    /// Return proof, commitment, and indices.
+    fn prove<const N: usize>(
+        params: &Params,
+        signature_list: &[IndividualSignature],
+        registration: &Registration,
+        msg: &[u8],
+        set_size: usize,
+    ) -> Option<Self> {
+        // Ensure that the commitment can be created
```
rrtoledo, line 39:
```suggestion
        // Ensure that the registration is closed
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let commitment = get_commitment::<N>(checksum, msg).to_vec();
+
+        // Collect valid individual signatures' byte representation into a hashmap
+        let valid_signatures = collect_valid_signatures::<N>(signature_list, registration, msg);
+
+        // Check if there are enough valid signatures
+        if valid_signatures.len() < set_size {
+            println!(
+                "Error: Not enough signatures! Expected at least {}, but got {}.",
```
rrtoledo, line 53:
```suggestion
                "Error: Not enough valid signatures to create an ATS! Expected at least {} signatures, but got {}.",
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+                set_size,
+                valid_signatures.len()
+            );
+            return None;
+        }
+
+        // Collect the byte representation of valid signatures into a Vec
+        let prover_set: Vec<Element> = valid_signatures.keys().copied().collect();
+
+        // Create the Alba proof using the prover set
```
rrtoledo, line 63:
```suggestion
        // Initialise Alba with the parameters and generates a proof using the prover set
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+            println!(
+                "Error: Not enough signatures! Expected at least {}, but got {}.",
+                set_size,
+                valid_signatures.len()
+            );
+            return None;
+        }
+
+        // Collect the byte representation of valid signatures into a Vec
+        let prover_set: Vec<Element> = valid_signatures.keys().copied().collect();
```
rrtoledo, line 61: Potentially stupid question: could we generate the proof directly on the indices?

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /// Verify AlbaThresholdSignature. Validate individual signatures and verify Alba proof.
+    fn verify<const N: usize>(
+        &self,
+        params: &Params,
+        registration: &Registration,
+        msg: &[u8],
+    ) -> bool {
+        // Check if the checksum exists in the registration
```
rrtoledo, line 89:
```suggestion
        // Ensure that the registration is closed, and retrieve the correct checksum
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+        let Some(checksum) = &registration.checksum else {
+            println!("Error: Registration is not closed.");
+            return false;
+        };
+
+        // Compute the commitment and compare with the stored commitment
```
rrtoledo, line 95: Since we compare this commitment anyway, we could remove it from the ATS proof!

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+            return false;
+        }
+
+        // Validate the individual signatures
+        if !validate_signatures(self, registration, &commitment) {
+            println!("Error: Signature validation failed.");
+            return false;
+        }
+
+        // Create the Alba proof and verify it
```
rrtoledo, line 108:
```suggestion
       // Initialise Alba with the parameters and generates a proof using the prover set
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+        };
+
+        // Compute the commitment and compare with the stored commitment
+        let commitment = get_commitment::<N>(checksum, msg).to_vec();
+        if commitment != self.commitment {
+            println!("Error: Commitment mismatch.");
+            return false;
+        }
+
+        // Validate the individual signatures
```
rrtoledo, line 102:
```suggestion
        // Aggregate the signatures and verify them at once
```

examples/aggregate_signature/helpers.rs
```diff
@@ -0,0 +1,102 @@
...
+    let aggregate_signature =
+        if let Ok(agg_sig) = AggregateSignature::aggregate(signature_refs.as_slice(), false) {
+            agg_sig.to_signature()
+        } else {
+            println!("Error: Failed to aggregate signatures.");
+            return false;
+        };
+
+    let public_key_refs: Vec<&PublicKey> = alba_threshold_signature
```
rrtoledo, line 75: I do not hink that you check that the indices correspond indeed to the verification keys!

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    // Register signers
+    for (index, signer) in signers.iter().enumerate() {
+        registration.register(signer.verification_key, index);
+    }
+    // Close registration to compute the checksum
+    registration.close::<DATA_LENGTH>();
+
+    // Filter out registered signers and create RegisteredSigner instances
+    let registered_signers: Vec<RegisteredSigner> = signers
```
rrtoledo, line 150: Why do you filter here the signers? Either they are all registered (honest scenario), or they are not (dishonest scenario) but we still want to check that the aggregator doesn't use these signatures.

rrtoledo (changes requested):
Overall looks good!

I am not sure if there is an error in the signature validation (it is late and I may have misread) as you may have not checked that the vf are indeed in the BTree. The registration also migth be problematic as the signer's index is not given by the registration.

Otherwise it looks good. I added some suggestions and a tiny optimizations (removing the checksum from the ATS proof).  I think you could simplify the code further by merging the two types of signers but that can be done in another PR.

examples/aggregate_signature/helpers.rs
curiecrypt: The `signature` only lives until the end of the iteration (from bytes iteration). So, I cannot use the ref I push for aggregation. Gives `borrowed value does not live long enough` error.

examples/README.md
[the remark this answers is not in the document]
curiecrypt: My idea was to provide a feature that allows a signer (`Signer` with just a signing key and a verification key) to register to sign different messages. It means that one signer can create more than one different `RegisteredSigner`s. Each registered signer is registered for a specific message.
Another point is that, we should also make sure that a signer generates only one signature. This could, of course, be checked by adding a boolean field. However, @rrtoledo and I are not sure whether it is the best way of doing this. WDYT @tolikzinovyev ?

examples/centralized_threshold.rs
curiecrypt: We might (?)

examples/README.md
```diff
@@ -0,0 +1,84 @@
...
+  - If the registration is closed and the key of the signer exists in the registered keys, `checksum` field of the signer is updated.
+
+### Registration
+- Registration includes registered keys as a `BTreeMap` and the checksum of all registered keys.
+- New registration is initialized by `new` function with an empty `BTreeMap`. Checksum is not initialized in this phase.
+- Registration process is closed with `close` function if the registration is not already closed. It is done by hashing all verification keys registered into the `checksum`.
+
+### Signature
+- A single signature is represented with `IndividualSignature`. 
+- It contains the bls signature and the index of the signer of signature.
```
rrtoledo, line 48:
```suggestion
- It contains the BLS signature and the index of the signer of signature.
```

examples/README.md
```diff
@@ -0,0 +1,84 @@
...
+  - Collects the valid signatures with `collect_valid_signatures`,
+  - If the length of valid signatures is less than the set size provided, abort the process,
+  - Creates the prover set by collecting only the signatures from the hash map of valid signatures,
+  - Creates the alba proof,
+  - Collects the registration indices of the elements that are in the alba proof by using valid signatures hashmap.
+  - Returns the alba proof, indices, and the commitment.
+- The `verify` function:
+  - Checks whether the registration is closed,
+  - Creates the commitment and compares it with the provided commitment,
+  - Verifies the signatures that creates the alba proof by calling `validate_signatures` function.
```
rrtoledo, line 83:
```suggestion
  - Verifies the signatures that are included in the alba proof by calling `validate_signatures` function.
```

examples/aggregate_signature/signer.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    /// Get closed registration. Update the registered signer with the checksum of registration.
+    pub(crate) fn get_closed_registration(&mut self, registration: &Registration) {
+        match &registration.checksum {
+            Some(checksum) => {
+                if registration.registered_keys.contains_key(&self.index) {
+                    self.checksum = Some(checksum.clone());
+                }
```
rrtoledo, line 69: You could add an else with an error message like "Error, Registration closed, signer not registered"

examples/aggregate_signature/signer.rs
```diff
@@ -0,0 +1,89 @@
...
+            }
+            None => {
+                println!("Error: Registration is not closed.");
+            }
+        }
+    }
+
+    /// Create an individual signature by signing `commitment = Hash(checksum || msg)`
+    pub(crate) fn sign<const N: usize>(&self, msg: &[u8]) -> Option<IndividualSignature> {
+        let commitment = self
```
rrtoledo, line 79: Not important. You could return None directly if checksum is None

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,372 @@
...
+        if let Some(checksum) = &registration.checksum {
+            let commitment = get_commitment::<N>(checksum, msg).to_vec();
+
+            if commitment != self.commitment {
+                println!("Error: Commitment mismatch.");
+                return false;
+            }
+
+            println!("-- Validating proof elements. ");
+            // Aggregate the signatures and verify them at once
```
rrtoledo, line 108:
```suggestion
            // Verify the signers were registered, aggregate the signatures and verify them at once
```

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,372 @@
...
+// Some basic tests
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rand_core::RngCore;
+
+    // Error: Registration is not closed. Cannot verify signatures.
+    // Number of collected valid signatures: 0.
+    #[test]
+    fn test_collect_valid_signatures_no_closed_reg() {
```
rrtoledo, line 227: Since this is a test, can you add an assert on the nb of valid sigs and print only if the test fails?

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,372 @@
...
+        println!(
+            "Number of collected valid signatures: {}.",
+            valid_signatures.len()
+        );
+    }
+
+    //Warning: No verification key found for index 999
+    // Expected 100 signatures, got 99.
+    #[test]
+    fn test_collect_valid_signatures_unregistered_signer() {
```
rrtoledo, line 270: Likewise (assert & print)

examples/centralized_threshold.rs
```diff
@@ -0,0 +1,372 @@
...
+        signer.register(&mut registration);
+
+        // Attempt to sign the message without a checksum and assert it returns None
+        assert!(signer.sign::<DATA_LENGTH>(&msg).is_none());
+    }
+
+    //Error: Registration is not closed.
+    // None
+    #[test]
+    fn test_get_checksum_before_closed() {
```
rrtoledo, line 324: Likewise (assert & print)

rrtoledo (commented):
LGTM.

I am not sure why we put tests here as this is an example but could you add asserts in each of these? We could add more tests, but let's do that in a disctint PR.

examples/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Workflow: ALBA Threshold Signature Example
+1. **Define Telescope Parameters**:
+  - Set the number of elements (`nb_elements`) to 1,000.
+  - Specify `soundness_param` and `completeness_param` as 128.0 each.
+  - Calculate the prover set size (`set_size`) as 80% of `nb_elements` and the lower bound as 20%.
+  - Initialize the Telescope with these parameters.
+2. **Initialize Signers**:
+  - Create a list of candidate signers (`Signer`s) with randomly generated keys.
+3. **Start Registration**:
```
djetchev, line 11: @curiecrypt :  It may be even a good idea to overcommunicate with a simple example the registration - e.g., a total of 3000 SPOs but only, say,1000 of them register and the Mithril lottery is done out of these 1000. WDYT?

djetchev (approved):
LGTM!

rrtoledo (approved):
LGTM

PR #120, opened 2025-01-10
Simple Example - Centralized Telescope with BLS 

curiecrypt:
## Content

This PR includes a basic example of integrating a multi-signature scheme into the Centralized Telescope. 
It consists of the functionality `KeyGen`, `Sign`, `Aggregate`, and `Verify` proposed in the _Definition 3._ of _Alba Definitions_ by @tolikzinovyev. 

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested

## Issue(s)
Relates to #114


Cargo.toml
```diff
@@ -10,6 +10,7 @@ include = ["**/*.rs", "Cargo.toml", "README.md", ".gitignore"]
 
 [dependencies]
 blake2 = "0.10.6"
+blst = "0.3.13"
```
rrtoledo, line 13: You can move this to dev-dependencies as it is used for tests and examples :)

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+struct ThresholdSignature {
+    proof: Proof,
+    key_list: Vec<PublicKey>,
+}
+
+impl AlbaSigner {
+    fn new(rng: &mut (impl RngCore + CryptoRng)) -> Self {
+        let mut ikm = [0u8; 32];
```
rrtoledo, line 27: What does `ikm` stands for? Can you rename it buffer if it is not standard instead?

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+use alba::centralized_telescope::proof::Proof;
+use alba::centralized_telescope::CentralizedTelescope;
+use blst::min_sig::{AggregatePublicKey, AggregateSignature, PublicKey, SecretKey, Signature};
+use blst::BLST_ERROR;
+use rand_chacha::ChaCha20Rng;
+use rand_core::{CryptoRng, RngCore, SeedableRng};
+use std::collections::HashMap;
+
+const DATA_LENGTH: usize = 48;
+pub(crate) type Element = [u8; DATA_LENGTH];
```
rrtoledo, line 13: Why is Element a pub crate?

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn sign<const N: usize>(&self, msg: &[u8]) -> [u8; N] {
+        let mut signature_to_byte = [0u8; N];
+        signature_to_byte.copy_from_slice(&self.signing_key.sign(msg, &[], &[]).to_bytes());
+        signature_to_byte
+    }
+}
+
+impl ThresholdSignature {
+    fn aggregate<const N: usize>(
+        alba_signatures: &HashMap<Element, usize>,
```
rrtoledo, line 47: It might be worth adding a comment on the type Element to say that it corresponds here to signatures (so that readers are not confused between elements and alba signatures).

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+            proof,
+            key_list: public_keys,
+        }
+    }
+
+    /// Validates individual signatures in the threshold signature
+    fn validate_signatures(&self, msg: &[u8]) -> bool {
+        let mut signatures = Vec::with_capacity(self.proof.element_sequence.len());
+        for sig_bytes in &self.proof.element_sequence {
+            if let Ok(signature) = Signature::from_bytes(sig_bytes.as_slice()) {
```
rrtoledo, line 74: I think you might be able to simplify the code by doing:

```
let Ok(signature) = Signature::from_bytes(sig_bytes.as_slice()) else { return false };
signatures.push(signature);
```

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+                signatures.push(signature);
+            } else {
+                println!("Error: Failed to parse signature from bytes.");
+                return false;
+            }
+        }
+        let signature_refs: Vec<&Signature> = signatures.iter().collect();
+        let aggregate_signature =
```
rrtoledo, line 82: Similarly
```
let Ok(signature) = AggregateSignature::aggregate(signature_refs.as_slice(), false) else { return false };
agg_sig.to_signature();
```

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+            if let Ok(agg_sig) = AggregateSignature::aggregate(signature_refs.as_slice(), false) {
+                agg_sig.to_signature()
+            } else {
+                println!("Error: Failed to aggregate signatures.");
+                return false;
+            };
+
+        let public_key_refs: Vec<&PublicKey> = self.key_list.iter().collect();
+        let aggregate_public_key =
```
rrtoledo, line 91: Similarly
```
let Ok(agg_pk) = AggregatePublicKey::aggregate(public_key_refs.as_slice(), false)
agg_pk.to_public_key();
```

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+        lower_bound: 20 * set_size / 100,
+    };
+
+    let mut key_list = HashMap::with_capacity(set_size as usize);
+    let mut signature_list = HashMap::with_capacity(set_size as usize);
+
+    for i in 0..set_size as usize {
+        let signer = AlbaSigner::new(&mut rng);
+        key_list.insert(i, signer.verification_key);
+        signature_list.insert(signer.sign::<DATA_LENGTH>(&msg), i);
```
rrtoledo, line 130: Why not having a single hashmap that uses a signature as key and public key as value? I think that would work.

rrtoledo (approved):
LGTM!

I think you can simplify a bit the code (replacing the if lets), optimize it a bit (you can use only one hashmap) and a couple of question on visibility and var name. :)

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    fn sign<const N: usize>(&self, msg: &[u8]) -> [u8; N] {
+        let mut signature_to_byte = [0u8; N];
+        signature_to_byte.copy_from_slice(&self.signing_key.sign(msg, &[], &[]).to_bytes());
+        signature_to_byte
+    }
+}
+
+impl ThresholdSignature {
```
tolikzinovyev, line 45: As before, I'm not in favor of this `impl` block. The same can be achieved without one. Here are problems I see.
1. Part of `verify()` function's input is in the struct and part is the regular function argument `msg`. This feels awkward.
2. If the member variables become public after moving the structs to a new file, then the struct will have public variables and an `impl` block. I have argued against it before.

I suggest following a style similar to what we have for alba in the main branch until that changes.
`ThresholdSignature`: similar to `Proof`, no `impl`
a "wrapper" struct `ThresholdSignatureScheme` similar to the alba wrapper struct having 3 functions: constructor `new()` accepting alba params and initializing alba, `aggregate()` creating `ThresholdSignature`, and `verify()` accepting `ThresholdSignature`.

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn aggregate<const N: usize>(
```
tolikzinovyev, line 46: Suggest replacing the hash maps with a slice of pairs (index, signature) like in my document. I think this would make the interface and implementation simpler. Particularly, there is no need for public keys since you can just attach the indices to complement the alba proof. So, in the language of my document, the signatures would be the data elements and the party indices would be the auxiliary data. Other options are:
1. data element (index, signature), no aux data
2. data element index, aux data signature

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn verify(&self, msg: &[u8], params: &Params) -> bool {
+        if self.validate_signatures(msg) {
+            let alba = CentralizedTelescope::create(params);
+            return alba.verify(&self.proof);
+        }
+        false
+    }
+}
+
+fn main() {
```
tolikzinovyev, line 112: Have you considered having a randomized test instead of the main function to also test correctness?

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+            } else {
+                println!("Error: Failed to aggregate public keys.");
+                return false;
+            };
+
+        let result = aggregate_signature.verify(false, msg, &[], &[], &aggregate_public_key, false);
+        result == BLST_ERROR::BLST_SUCCESS
+    }
+
+    fn verify(&self, msg: &[u8], params: &Params) -> bool {
```
tolikzinovyev, line 103: This needs the original party public keys as input. Otherwise, the public keys can be forged. Suggest passing a slice of public keys (one for each party) like in my document.

examples/simple_example.rs
tolikzinovyev, line 130: let's pass only n_p signatures to aggregate() to be more realistic

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+        let sk = SecretKey::key_gen(&ikm, &[])
+            .expect("Error occurs when the length of ikm < 32. This will not happen here.");
+        let pk: PublicKey = sk.sk_to_pk();
+        Self {
+            signing_key: sk,
+            verification_key: pk,
+        }
+    }
+
+    fn sign<const N: usize>(&self, msg: &[u8]) -> [u8; N] {
```
tolikzinovyev, line 38: I don't think we need generics here or elsewhere.

tolikzinovyev (commented):
Very good start. One request: let's not commit to the main branch directly. If the element size needs to be changed, it can be done in this PR.

examples/simple_example.rs
tolikzinovyev: I think we should use the signature type directly.

examples/simple_example.rs
curiecrypt: In order to make this code work, we need to define another variable to hold the signature, which is created with `to_signature()`. Like following: 
```
let Ok(aggregate_signature) =
            AggregateSignature::aggregate(signature_refs.as_slice(), false)
        else {
            return false;
        };
let final_signature = aggregate_signature.to_signature();
```
Because the function `to_signature()` takes an `AggregateSignature` as input and returns a `Signature`.

examples/simple_example.rs
curiecrypt: Similarly.

examples/simple_example.rs
rrtoledo: I was more curious than anything :)

examples/simple_example.rs
curiecrypt: Yes, I will add them.

examples/simple_example.rs
rrtoledo: Sorry, I wrongly thought that to_signature was an in place function I understand now!
It just becomes a matter of coding style then which is not very important. I will resolve this comment now and leave you to choose which style you prefer.

examples/simple_example.rs
curiecrypt: This is the style we all (@djetchev, @rrtoledo ) agreed on a month ago.
Also, it will be adopted in the src code; @rrtoledo has been working on it lately. So, for consistency, it makes more sense to adopt this style for this example.

examples/simple_example.rs
```diff
@@ -0,0 +1,136 @@
...
+use blst::min_sig::{AggregatePublicKey, AggregateSignature, PublicKey, SecretKey, Signature};
+use blst::BLST_ERROR;
+use rand_chacha::ChaCha20Rng;
+use rand_core::{CryptoRng, RngCore, SeedableRng};
+use std::collections::HashMap;
+
+const DATA_LENGTH: usize = 48;
+pub(crate) type Element = [u8; DATA_LENGTH];
+
+struct AlbaSigner {
```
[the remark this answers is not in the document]
rrtoledo: I would disagree. I think there is merit in keeping such a simple example in a single self contained file. It is only 100 lines long so it is not very difficult to keep track of everything here.

examples/simple_example.rs
rrtoledo: As @curiecrypt said, we agreed to change the coding style to impl around a month ago. So this should not be an issue.

As for the structure part, I suggest to go for the standard interface `signature.verify(param, vk(?s), msg)` which is adopted by Mithril and other projects. The `struct ThresholdSignature` would hence becomes an alias only, which seems great in our case: showing in an example how to use Alba.

examples/simple_example.rs
rrtoledo: Whichever structure (hashmaps, vectors, a vector of tuples...) you want to use is fine to me.
I would not recommend using indices in the proof however, as we want this to be the most straightforward example.

examples/simple_example.rs
rrtoledo: You could write show in 3 blocks that providing np and more than np sigs will produce a threshold signature, less than np won't however.

examples/simple_example.rs
curiecrypt: Copy-paste left-over. Removed!

examples/simple_example.rs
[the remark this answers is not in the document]
tolikzinovyev: I think there is merit in showing that any unique signature scheme can be used in place of BLS. So maybe individual verification + a comment saying that for BLS you can also aggregate for efficiency?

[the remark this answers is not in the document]
rrtoledo: Sounds good to me too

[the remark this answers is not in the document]
curiecrypt: Agreed!

examples/simple_example.rs
curiecrypt: Update: I decided not to add any randomized tests for this example.
- There is nothing to test on the side of the multi-signature scheme, we don't even check the validity of the signatures before aggregation.
- The randomized tests for the alba side should be handled in its source folder. 
So, except for showing a basic flow, I don't see a case to test here.

tolikzinovyev: There is logic in the aggregate and verify functions that is currently untested. But if you don't want to add tests, that's fine.

examples/simple_threshold_signature/threshold_signature.rs
```diff
@@ -0,0 +1,89 @@
...
+        let prover_set = signatures
+            .iter()
+            .map(|s| s.signature.to_bytes())
+            .collect::<Vec<Element>>();
+        println!("-- Creating alba proof. ");
+        let proof = alba.prove(&prover_set).unwrap();
+        println!("-- Alba proof created: ");
+        println!(
+            " - Numbers of retries done to find the proof: {}",
+            proof.retry_counter
```
rrtoledo, line 27: Should we implement getters for the proof and other structs instead of having public fields (like for Telescope)? (We would do it in another PR of course)

examples/simple_example.rs
```diff
@@ -0,0 +1,64 @@
...
+        signature_list.push(signer.sign(&msg, i));
+    }
+    println!("--------------------------------------------------------");
+    println!(" -- {set_size} (sk, pk) are generated.");
+    println!(" -- {set_size} signatures are generated.");
+
+    println!("--------------------------------------------------------");
+    println!("----------- Generating Alba multi-signature. -----------");
+    let (threshold_signature, public_keys) =
+        ThresholdSignature::aggregate(&signature_list, &alba, &key_list);
```
tolikzinovyev, line 52: This still doesn't show that n_p signatures is enough. What's the reason not to pass only n_p signatures?

examples/simple_threshold_signature/signature.rs
```diff
@@ -0,0 +1,10 @@
+use blst::min_sig::Signature as BLSSignature;
```
tolikzinovyev, line 1: Suggest `BlsSignature` for readability

examples/simple_threshold_signature/threshold_signature.rs
```diff
@@ -0,0 +1,89 @@
...
+            }
+        }
+        true
+    }
+
+    pub(crate) fn verify(&self, msg: &[u8], alba: &Telescope, public_keys: &[PublicKey]) -> bool {
+        if self.validate_signatures(msg, public_keys) {
+            return alba.verify(&self.proof);
+        }
+        false
```
tolikzinovyev, line 87:
```suggestion
        self.validate_signatures(msg, public_keys) && alba.verify(&self.proof)
```

examples/simple_example.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    println!("-- Alba multi-signature is generated.");
+    println!("--------------------------------------------------------");
+    println!("----------- Verifying Alba multi-signature. ------------");
+
+    if threshold_signature.verify(&msg, &alba, &public_keys) {
```
tolikzinovyev, line 58: shouldn't the list of all public keys be in the input?

examples/simple_example.rs
curiecrypt: It is already `n_p` signatures. Using the new structure, I calculate `set_size = n_p` as follows:
image

curiecrypt: Do you mean something different?

examples/simple_threshold_signature/threshold_signature.rs
curiecrypt: Let's do it in another PR 👍

examples/simple_example.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    println!(" - Soundness parameter: {soundness_param}");
+    println!(" - Completeness parameter: {completeness_param}");
+    println!(" - Prover set size: {set_size}");
+    println!(" - Lower bound: {lower_bound}");
+
+    // Create the telescope structure
+    let alba = Telescope::create(soundness_param, completeness_param, set_size, lower_bound);
+
+    let mut public_key_list: Vec<(usize, PublicKey)> = Vec::with_capacity(nb_elements as usize);
```
tolikzinovyev, line 39: The `usize` variable always contains its own index, so it is redundant and can be removed.

examples/simple_threshold_signature/threshold_signature.rs
```diff
@@ -0,0 +1,114 @@
...
+            .filter_map(|element| BlsSignature::from_bytes(element).ok())
+            .collect();
+
+        // Collect the indices of the signatures that create alba proof.
+        let mut indices = Vec::with_capacity(proof_signatures.len());
+        for sig in &proof_signatures {
+            if let Some(signature_entry) = signatures.iter().find(|entry| entry.signature == *sig) {
+                if public_key_list
+                    .iter()
+                    .any(|entry| entry.0 == signature_entry.index)
```
tolikzinovyev, line 57: I think this logic wouldn't work if you pass signatures to this function that are not index 0, 1, ..., n_p-1. That's why in my document Aggregate gets triples (i, s_i, w_i). In particular, the index i of the signature is included.

Also, `public_key_list` seems redundant in this function.

PR #137, opened 2025-01-28
Hash as input

### LINKED ISSUE #36: User provided hash function
## Why
Support user provided hash functions so that they can address domain separation when using Alba certificates several times in the same application.

rrtoledo:
## Content

This PR aims at updating protocols to support different hashes, input by the user.

## Pre-submit checklist

- Branch
- PR
    - [x] Self-reviewed the diff
    - [x] Reviewer requested
- Documentation

## Comments

I have kept Blake2bVar when generating data, but changed to Sha256 in tests, benches and examples as it is faster on my machine.

I also fixed some old Clippy warnings.

## Issue(s)

Relates to #36


src/centralized_telescope/proof.rs
```diff
@@ -248,21 +275,25 @@ impl Proof {
...
-        hasher.update(r.hash);
-        let digest: Hash = hasher.finalize().into();
-        sample::sample_bernoulli(&digest, valid_proof_probability)
+    fn proof_hash(valid_proof_probability: f64, r: &Round<H>) -> bool {
+        let digest = H::new()
+            .chain_update(b"Telescope-proof_hash")
+            .chain_update(r.hash)
+            .finalize()
+            .to_vec();
+        let hash = truncate(&digest);
```
tolikzinovyev, line 296: Looks like that old method doesn't work since `GenericArray` only implements `From` for certain array sizes. https://docs.rs/generic-array/0.14.7/src/generic_array/impls.rs.html#183-216

At the same time truncation is sketchy, the hash function could have a very small output in principle. I think a better solution, though not ideal, would be to replace `.to_vec()` with `.to_slice()`, then runtime assert that the slice size is at least 16 bytes and pass an array of 16 bytes to sample.

tests/common/mod.rs
```diff
@@ -1,17 +1,19 @@
 //! Test & Bench helpers
 
-use blake2::digest::{Update, VariableOutput};
-use blake2::Blake2bVar;
+use digest::VariableOutput;
 
 /// Generate a set of items given the set size and a seed
 /// Items are generated by hashing the current index
-pub(crate) fn gen_items<const N: usize>(seed: &[u8], set_size: u64) -> Vec<[u8; N]> {
+pub(crate) fn gen_items<H: VariableOutput, const N: usize>(
```
tolikzinovyev, line 7: Does this need a generic hash function?

tests/centralized_telescope.rs
```diff
@@ -2,15 +2,19 @@
...
 mod common;
 use common::gen_items;
 
+use blake2::Blake2bVar;
+use sha2::Sha256;
+
 const DATA_LENGTH: usize = 48;
 
-fn test(created_with_params: bool) {
+fn test<H: Digest + FixedOutput>(created_with_params: bool) {
```
tolikzinovyev, line 17: Not sure generics are needed here.

src/centralized_telescope/proof.rs
```diff
@@ -6,22 +6,25 @@ use super::params::Params;
...
-pub struct Proof {
+pub struct Proof<H: Digest + FixedOutput> {
     /// Numbers of retries done to find the proof
     pub retry_counter: u64,
     /// Index of the searched subtree to find the proof
     pub search_counter: u64,
     /// Sequence of elements from prover's set
     pub element_sequence: Vec<Element>,
+    // Phantom type to link the tree with its hasher
+    hasher: PhantomData<H>,
```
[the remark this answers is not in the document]
rrtoledo: This was done similarly to Mithril code base to keep a link between a proof and the underlying hash function.
Of course, in practice this is 0 byte but it keeps the code safe when compiling.

src/centralized_telescope/proof.rs
rrtoledo: Good call!

tests/centralized_telescope.rs
rrtoledo: Sure we can remove it

curiecrypt (dismissed):
LGTM

src/centralized_telescope/telescope.rs
```diff
@@ -130,15 +131,16 @@ impl Telescope {
...
     /// let telescope = Telescope::create(64.0, 64.0, set_size, 100);
     /// let mut prover_set = Vec::new();
     /// for i in 0..set_size {
     ///     prover_set.push([(i % 256) as u8 ; 48]);
     /// }
-    /// let proof = telescope.prove(&prover_set).unwrap();
+    /// let proof = telescope.prove::<Sha256>(&prover_set).unwrap();
     /// ```
-    pub fn prove(&self, prover_set: &[Element]) -> Option<Proof> {
+    pub fn prove<H: Digest + FixedOutput>(&self, prover_set: &[Element]) -> Option<Proof<H>> {
```
tolikzinovyev, line 143: Thinking about this PR again, having domain separation in mind per design document. Two points.

1. I think we and the user would benefit from an example showing how to construct and use a hash function that has a domain id inside. Personally, right now I don't understand how easy it is to do.
2. The current interfaces would work if the user has a finite number of domain ids since they can be defined directly in the hash function type. However, if we don't know at compile time how many possible domain ids there are, the current interface will not work. The user would need to provide not just the hash function type but also a hash function object.

src/utils/types.rs
```diff
@@ -7,4 +7,12 @@ pub(crate) type Element = [u8; DATA_LENGTH];
 pub(crate) const DIGEST_SIZE: usize = 32;
 
 /// Hash type for internal hashes
-pub(crate) type Hash = [u8; DIGEST_SIZE];
+pub type Hash = [u8; DIGEST_SIZE];
+
+pub(crate) fn truncate(data: &[u8]) -> Hash {
+    debug_assert!(data.len() >= 16);
```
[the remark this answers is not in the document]
rrtoledo: You were the one who asked for 16 bytes before instead of DIGEST_SIZE but I wasn't really sure why. Wouldn't it be better to just fix everything to DIGEST_SIZE?

examples/centralized_threshold.rs
```diff
@@ -14,18 +14,20 @@ use std::time::Instant;
 mod aggregate_signature;
 const DATA_LENGTH: usize = 48;
 pub(crate) type Element = [u8; DATA_LENGTH];
+use digest::{Digest, FixedOutput};
+use sha2::Sha512;
```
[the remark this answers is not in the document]
rrtoledo: No particular reasons. We can use either. As this is an example it shouldn't run in the tests.

src/centralized_telescope/telescope.rs
rrtoledo: I had a go at this and it is not simple indeed (as you need to deal with `OutputSizeUser` and GenericArrays).
As this is not a strict requirement for this PR, I'll open an issue to provide a hash example.

src/utils/types.rs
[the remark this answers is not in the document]
tolikzinovyev: Yes. Only suggesting to set `DIGEST_SIZE` to 16.

curiecrypt (approved):

